PR #44, opened 2025-02-14
feat: project card design implemented

serhatkaya:


indigo-frontend/src/core/components/project/project-item/project-item.component.html
```diff
@@ -0,0 +1,71 @@
...
+      <div>Last Edited</div>
+      <div>14 Oct 2024 by Emmy Cooper</div>
+    </div>
+    <div class="flex items-center justify-between">
+      <div>Created</div>
+      <div>08 Oct 2024 by John Doe</div>
+    </div>
+  </div>
+  <div class="flex items-center mt-4">
+    <div class="flex -space-x-2">
```
aamialkovich, line 59: Let's implement this functionality in dynamic way, so it will display number of hidden users based on users list. 😃

PR #89, opened 2025-03-04
feat: added accordion component

kalyani-epam:


indigo-frontend/src/core/components/common/accordion/accordion.component.scss
```diff
@@ -0,0 +1,50 @@
+.example-accordion {
```
aamialkovich, line 1: Could you please get rid of _example_ prefix as it is common component.

ui/package-lock.json
aamialkovich, line 1: Seems it is not needed as package.json wasn't updated.

aamialkovich (changes requested):
Could you please implement it in way so it will be used somehow:

```
<app-accordion [title]="'Experiment description'">
  <div>Some content here</div>
</app-accordion>
```

indigo-frontend/src/core/components/common/accordion/accordion.component.scss
serhatkaya: I believe we should focus on Tailwind-based development to maintain a consistent codebase.

indigo-frontend/src/core/components/common/accordion/accordion.component.html
```diff
@@ -0,0 +1,22 @@
...
+      [attr.aria-expanded]="accordionItem.expanded" [attr.aria-controls]="'accordion-body-' + index">
+      {{ item.title }}
+      <span class="text-sm text-gray-600">
+        <em [class.indicon-chevron_up]="!accordionItem.expanded"
+          [class.indicon-chevron_down]="accordionItem.expanded"></em>
+      </span>
+    </button>
+    @if(accordionItem.expanded) {
+    <div class="p-4" role="region" [style.display]="accordionItem.expanded ? '' : 'none'"
+      [attr.id]="'accordion-body-' + index" [attr.aria-labelledby]="'accordion-header-' + index" [innerHTML]="item.description">
```
aamialkovich, line 17: Please get rid of innerHTML and use angular content projection like in modal.component

indigo-frontend/src/core/components/common/accordion/accordion.component.ts
```diff
@@ -0,0 +1,14 @@
+import { Component } from '@angular/core';
+import {CdkAccordionModule} from '@angular/cdk/accordion';
+
+@Component({
+  selector: 'app-accordion',
+  imports: [CdkAccordionModule],
+  templateUrl: './accordion.component.html'
+})
+export class AccordionComponent {
+  items = [{title: 'Item 1', description: '<table><thead><tr><th>Header 1</th><th>Header 2</th><th>Header 3</th></tr></thead><tbody><tr><td>Data 1</td><td>Data 2</td><td>Data 3</td></tr></tbody></table>'},
```
aamialkovich, line 10: No need to put items here. As this is single accordion component, we can rely only on title as angular input here and content which will be passed to this accordion.

kalyani-epam: Removed items from Accordion Component

indigo-frontend/src/core/components/common/accordion/accordion.component.html
kalyani-epam: Removed innerHtml and added ngContent

PR #90, opened 2025-03-05
Input component

### LINKED ISSUE #74: [FE] Implement input control
- Component should be implemented by using ControlValueAccessor

Image

eu-peregudova:
Closes #74 

- Added input component

image

### DIFF
--- a/indigo-frontend/src/core/components/common/input/input.component.html
+++ b/indigo-frontend/src/core/components/common/input/input.component.html
@@ -1,0 +1,40 @@
+<label class="flex flex-col gap-2">
+  <span
+    class="text-sm font-medium"
+    [class]="{
+      'text-red-200': hasError(),
+    }"
+  >
+    {{ label() }}
+  </span>
+  @if (required()) {
+    <span class="text-red-200">*</span>
+  }
+
+  <div class="relative flex items-center">
+    <input
+      [class]="
+        {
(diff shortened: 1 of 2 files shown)


indigo-frontend/src/core/components/common/input/input.component.ts
```diff
@@ -0,0 +1,66 @@
...
+    ],
+  imports: [ClassPickerPipe, CommonModule],
+})
+export class InputComponent implements ControlValueAccessor {
+  label = input.required<string>();
+  placeholder = input<string>();
+  required = input<boolean>(false);
+  disabled = input<boolean>(false);
+  hasError = input<boolean>(false);
+  initialValue = input<string>('');
```
aamialkovich, line 25: Do we need initialValue? Seems that `value` is alreadly enough.

indigo-frontend/src/core/components/common/input/input.component.html
```diff
@@ -0,0 +1,28 @@
+<label class="flex flex-col gap-2">
```
aamialkovich, line 1: Could you please check formatting with prettier? Seems it has some issues.

indigo-frontend/src/core/components/common/input/input.component.ts
eu-peregudova: Since I am using signal input and it is immutable, can't have both input and current changeable value in one

indigo-frontend/src/core/components/common/input/input.component.html
eu-peregudova: Fixed that

aamialkovich (approved):

Issue #93, opened 2025-03-05
[FE] Implement api service


aamialkovich:
In scope of this task we need to implement service which will be wrap @angular/http methods.

PR #98, opened 2025-03-11
#93 added api service

AlxSEPAM:


indigo-frontend/src/core/services/api.service.ts
aamialkovich, line 1: Just a small proposal. Let's rename `data` variable to `body` as it is more common for http requests.

aamialkovich (approved):

PR #101, opened 2025-03-14
Suffix component

### LINKED ISSUE #91: [FE] Implement input control with suffix
https://www.figma.com/design/KjAepepwNqfSv3aGD75S6r/%F0%9F%8C%88-Indigo-UI-KIT?node-id=193-1270&t=XilPmDhRtuNWWsxP-0

eu-peregudova:
- Used input component and select component to create suffix component
- Updated styles in both components for this case
- Refactored input component's handling of disabled state

image

### DIFF
--- a/indigo-frontend/src/core/components/common/input-suffix/input-suffix.component.ts
+++ b/indigo-frontend/src/core/components/common/input-suffix/input-suffix.component.ts
@@ -32,7 +32,7 @@ import { InputSuffixValue } from '../../../types/input-suffix.i';
 })
 export class InputSuffixComponent implements ControlValueAccessor {
   label = input.required<string>();
-  placeholder = input<[string, string]>();
+  placeholder = input<InputSuffixValue>();
   required = input<boolean>(false);
   hasError = input<boolean>(false);
   items = input<DropdownMenuItem[]>([]);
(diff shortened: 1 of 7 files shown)


indigo-frontend/src/core/components/common/input-suffix/input-suffix.component.ts
```diff
@@ -0,0 +1,91 @@
...
+  ],
+})
+export class InputSuffixComponent implements ControlValueAccessor {
+  label = input.required<string>();
+  placeholder = input<[string, string]>();
+  required = input<boolean>(false);
+  hasError = input<boolean>(false);
+  items = input<DropdownMenuItem[]>([]);
+  valueChange = output<[string, string]>();
+  value = signal<[string, string]>(['', '']);
```
aamialkovich: Maybe it will be more readable to use object instead of array?
Something like `{ suffix: string; value: string }`

eu-peregudova: Agree, updated that part

indigo-frontend/src/core/components/common/input-suffix/input-suffix.component.ts
```diff
@@ -0,0 +1,95 @@
...
+  imports: [
+    SelectComponent,
+    InputComponent,
+    ClassPickerPipe,
+    ReactiveFormsModule,
+  ],
+})
+export class InputSuffixComponent implements ControlValueAccessor {
+  label = input.required<string>();
+  placeholder = input<[string, string]>();
```
aamialkovich: Sorry, forgot to add the same comment as for value. I think we can use object here too.

aamialkovich (approved):

indigo-frontend/src/core/components/common/input-suffix/input-suffix.component.ts
eu-peregudova: Makes sense, actually!

PR #104, opened 2025-03-17
Feature/103 project list

### LINKED ISSUE #103: [FE] Projects backend integration
In scope of this task we should replace mock data with backend integrated.

[Swagger](https://indigo-eln-dev.test.lifescience.opensource.epam.com/swagger/eln/#/ELN%20Resource/get_api_eln_projects)

serhatkaya:
closes #103


indigo-frontend/src/core/utils/http.util.ts
```diff
@@ -0,0 +1,18 @@
...
+  filter: Record<string, unknown>,
+  paging?: PagedRequest,
+) {
+  let params = new HttpParams();
+
+  const filters = { ...filter, ...paging };
+
+  Object.keys(filters).forEach(
+    (key) =>
+      Boolean(filters[key]) && (params = params.append(key, filters[key])),
```
aamialkovich: Maybe to use !! instead of Boolean constructor as MDN doesn't recommend to use such expression

indigo-frontend/src/core/types/entities/project.i.ts
```diff
@@ -0,0 +1,14 @@
+import { BaseEntity } from './base-entity.i';
+
+export interface Project extends BaseEntity {
+  name: string;
+  notebookCount: number;
+  experimentCount: ExperimentCount;
+}
+
+export interface ExperimentCount {
```
aamialkovich: Could you please clarify with BE side if it is a correct interface. It is presented in Swagger but I'm not sure we need it 😄

indigo-frontend/src/core/services/api.service.ts
aamialkovich, line 1: Could you also add method for adding entity via post request?
And it seems that unit tests for api service are not valid anymore.

aamialkovich (approved):

indigo-frontend/src/core/types/entities/project.i.ts
serhatkaya: I wasn't certain about it either thats why i added 
  [key: string]: number;
  
  :D 
  i'll align with B.E

Issue #154, opened 2025-06-16
[BE] Filtering and Sorting in Project List


denis-dzenskevich:
The User optionally sorts Projects (not in current Release)

- The User selects a sorting option:
  - Earliest
  - Latest
- The IndigoELN System sorts the projects based on the selected criteria.
- The User views the updated project list in the selected order.

The User optionally views only Projects created by the User

- The User selects the option "My Entities".
- The IndigoELN System filters the project list to show only those created by the User.
- The User views the updated project list.

PR #209, opened 2025-08-01
154 project list filtering

m-markau:
Add Filtering and Sorting in Project List (#154)


m-markau:
Wrong branch

denis-dzenskevich (changes requested):
All good, but can we make an enum for sorting (e. g. enum ProjectsSortOrder)? It will be simpler to tell what options are available and more convenient to use in code

backend/eln/eln-api/src/main/java/com/epam/indigoeln/eln/api/ProjectAPI.java
```diff
@@ -20,7 +20,9 @@ public interface ProjectAPI extends BaseAPI {
 
     @GET
     @Path("/projects")
-    Page<ProjectDTO> getProjects(@QueryParam("search") @Nullable String search, @BeanParam Paging paging);
+    Page<ProjectDTO> getProjects(@QueryParam("search") @Nullable String search, @QueryParam("sort") @Nullable String sort,
```
denis-dzenskevich: Please use enum here too

denis-dzenskevich (changes requested):

backend/eln/eln-api/src/main/java/com/epam/indigoeln/eln/api/ProjectAPI.java
m-markau: Done

backend/eln/eln-core/src/main/java/com/epam/indigoeln/eln/repository/ProjectRepository.java
```diff
@@ -25,12 +23,21 @@ public ProjectRepository() {
         super(EntityType.PROJECT);
     }
 
-    public ListWithTotal<ProjectDTO> findAll(@Nullable String search, Paging paging) {
+    public ListWithTotal<ProjectDTO> findAll(@Nullable String search, @Nullable SortOrder sort, @Nullable UserEntity createdByUser, Paging paging) {
+        Sort panacheSort = switch (sort) {
```
denis-dzenskevich, line 27: I guess it will be unexpected for the users if default sorting doesn't match any of options they can select

Maybe we will always sort by modifiedAt (asc/desc) instead of createdAt? Also, we won't need DEFAULT enum branch in this case - we'll just have one of the options as default

backend/eln/eln-api/src/main/java/com/epam/indigoeln/eln/model/SortOrder.java
```diff
@@ -0,0 +1,16 @@
...
+
+import lombok.Getter;
+
+@Getter
+public enum SortOrder {
+    EARLIEST("Earliest"),
+    LATEST("Latest"),
+    DEFAULT("Default");
+
+    private final String value;
```
denis-dzenskevich: Never used anymore?

denis-dzenskevich (changes requested):

backend/eln/eln-core/src/main/java/com/epam/indigoeln/eln/repository/ProjectRepository.java
m-markau: Just to clarify, are you asking about changing the behavior of the default sorting?

Before this PR, the default sorting was managed by the DEFAULT_SORT variable in the BaseRepository class:

Sort DEFAULT_SORT = io.quarkus.panache.common.Sort.descending("modifiedAt");

In this PR, I added a default value in SortOrder, which is mimic the same behavior as the previous 
implementation.

denis-dzenskevich: I mean we can make LATEST/EARLIEST sort by modifiedAt, not by createdAt (it's not really specified, I think we are free to do so)

So there will be only 2 options:
- SortOrder.LATEST (meaning modifiedAt desc)
- SortOrder.EARLIEST (meaning modifiedAt asc)
- 
In this way, LATEST will be the same as DEFAULT_SORT, and we can use LATEST by default if UI didn't send sort parameter explicitly

denis-dzenskevich (approved):

Issue #208, opened 2025-07-31
[BE] Create template


denis-dzenskevich:
- add tabs
- add validations
- add missing component types

PR #275, opened 2025-09-09
#208: Create template

m-markau:
Implementation for task #208


backend/eln/eln-core/src/main/java/com/epam/indigoeln/eln/util/TemplateValidationUtil.java
```diff
@@ -0,0 +1,62 @@
...
+public class TemplateValidationUtil {
+
+    /**
+     * Validates a template request based on the provided specification.
+     *
+     * @param templateRequest    the template request to validate.
+     * @param templateRepository the template repository.
+     */
+    public static void validateTemplateRequest(TemplateRequest templateRequest, TemplateRepository templateRepository) {
+        if (templateRequest.getName().isBlank()) {
```
denis-dzenskevich: Shall be handled automatically if you put @NotNull on a field and add @Valid on a Controller (Resource) parameter. Same for other empty/blank checks

See e. g. ExperimentServiceTest.testCreateExperimentValidation

backend/eln/eln-core/src/main/java/com/epam/indigoeln/eln/service/TemplateService.java
```diff
@@ -47,10 +54,22 @@ public TemplateDetailsDTO getTemplate(UUID templateId) {
...
     public TemplateDetailsDTO editTemplate(UUID templateId, TemplateEditRequest request) {
         aclService.ensureTopLevelAccess(ApplicationPermission.MANAGE_TEMPLATES);
+
         TemplateEntity template = templateRepository.get(templateId);
-        editProperty(request.getName(), template::setName);
+
+        // Edit the template name if provided
+        editProperty(request.getName(), name -> {
+            // Validate the new name for uniqueness
+            if (templateRepository.existsByName(name)) {
```
denis-dzenskevich: Optional, but I would rely on a database constraint for that (like in ProjectService.createProject for example)

First, it will work uniformly for new and updated entities. Second, it saves from corner cases. E. g. update with the same name will trigger an error here

backend/eln/eln-core/src/test/resources/com.epam.indigoeln.eln.entity/template_entity.json
```diff
@@ -0,0 +1,67 @@
+{
```
denis-dzenskevich: Do we need this file in Git? If it's just to see the output, please generate file somewhere inside "build" directory so it's not included

denis-dzenskevich (changes requested):

backend/eln/eln-core/src/main/java/com/epam/indigoeln/eln/util/TemplateValidationUtil.java
m-markau: Done

backend/eln/eln-core/src/main/java/com/epam/indigoeln/eln/service/TemplateService.java
m-markau: Done. Rollback to original version

backend/eln/eln-core/src/test/resources/com.epam.indigoeln.eln.entity/template_entity.json
m-markau: I put in build directory

backend/eln/eln-core/src/main/java/com/epam/indigoeln/eln/service/TemplateService.java
denis-dzenskevich: Seems it's still checked by existsByName?

m-markau: I verified it ones more time. 

In latest version of this branch: backend-v3-208-create-template - method rollbacked.

Here is what I see in git:
    public TemplateDetailsDTO editTemplate(UUID templateId, TemplateEditRequest request) {
        aclService.ensureTopLevelAccess(ApplicationPermission.MANAGE_TEMPLATES);

        TemplateEntity template = templateRepository.get(templateId);

        editProperty(request.getName(), template::setName);

        ModelUtil.updateDates(template, userService.getCurrentUser());

        templateRepository.flushAndClear();

        return getTemplate(templateId);
    }

No check for existsByName

m-markau: I have replaced TemplateRepository.existsByName() by DB unique constraint on "name" field

denis-dzenskevich (approved):

PR #301, opened 2025-09-30
Merge updated report into the backend-v3 branch

andrzej16913:
The report generating code was updated to make it more similar to the old Indigo reports. Code was edited in report service, report service tests and in report jrxml files.

### DIFF
--- a/backend/reports/reports-core/src/main/jasperreports/ExperimentReport.jrxml
+++ b/backend/reports/reports-core/src/main/jasperreports/ExperimentReport.jrxml
@@ -164,7 +164,7 @@
 				<text><![CDATA[REACTION DETAILS]]></text>
 			</element>
 			<element kind="staticText" uuid="2eba9eca-c1c6-4f57-8a7a-7c5ea729b4ab" stretchType="ContainerHeight" x="0" y="24" width="141" height="20" bold="true">
-				<text><![CDATA[Project Name]]></text>
+				<text><![CDATA[Created Date]]></text>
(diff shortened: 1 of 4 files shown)


backend/reports/reports-core/src/main/jasperreports/ExperimentReport.jrxml
```diff
@@ -176,7 +176,7 @@
 				</box>
 			</element>
 			<element kind="textField" uuid="3926cfa2-01c8-4d04-be7e-d6cc4519ff2c" stretchType="ContainerHeight" x="141" y="24" width="138" height="20" textAdjust="StretchHeight">
-				<expression><![CDATA[$F{project}.getName()]]></expression>
+				<expression><![CDATA[$F{experiment}.getCreatedAt().format(java.time.format.DateTimeFormatter.ofPattern("MMM d, u HH:mm:ss", Locale.ENGLISH)) + " UTC"]]></expression>
```
denis-dzenskevich, line 179: Please make UTC part of format pattern. ~~For now, I guess our containers run with UTC timezone anyway, but ideally, we'd configure time zone for server or per user, and it will be formatted correctly~~

Actually, can we make time zone configurable? You can set default timezone (e.g. UTC) in microprofile-config.properties in reports-core, and we will be able to override it when deploying. In the report service, you can inject its value as @ConfigProperty and use in the code, converting all dates to that time zone

denis-dzenskevich (changes requested):

denis-dzenskevich (commented):
Also, can you please attach the generated report here?

Issue #304, opened 2025-09-30
[BUG] My entities toggle does not work


Roman888Kuritsyn:
The User navigates to the Dashboard page.
The User selects the option "My Entities".
The IndigoELN System filters the project list to show only those created by the User.
**Actual result:**
Nothing changes
**Expected result:**
The User views the updated project list.saving.
Attributes are displayed
**Related tasks:**
#110

nadzeyapiatrouskaya-boop:
This issue is still being reproduced.
As is: 
A user-created project appears in the general list, but it doesn't appear when switching to My Entities. It also disappears from the list when switching back to the general list of projects. After refreshing the page, the created project appears correctly in the general list of projects.
_**NB! Reproduce on Notebooks and Experiment levels**_
Please see attached video: 

https://github.com/user-attachments/assets/f5df2f4f-865e-4d24-bba7-ae9c19fc3b86

nadzeyapiatrouskaya-boop:
The bug has been fixed correctly. Verified on Dev environment 2026-02-01
In confirmation:

https://github.com/user-attachments/assets/0ffd235e-d0a4-4e2c-94d5-39875d97f74c

PR #324, opened 2025-10-15
#304 Added "My Entities" filter functionality

gunabhushan:
#304 Added "My Entities" filter functionality

### DIFF
--- a/indigo-frontend/src/app/pages/project/project-list/project-list.component.ts
+++ b/indigo-frontend/src/app/pages/project/project-list/project-list.component.ts
@@ -108,6 +108,8 @@ export class ProjectListComponent
   }
 
   onMyEntitiesOnlyChange(value: boolean) {
-    this.myEntities(value);
+    this.filters['createdByMe'] = value;
+    this.dataBh.next([]);
+    this.fetchDataAndUpdateQueryParams();
   }
 }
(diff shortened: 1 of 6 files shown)


indigo-frontend/src/core/components/util/paginated.base.ts
```diff
@@ -232,6 +232,11 @@ export abstract class PaginatedBase<T> {
     this.fetchDataAndUpdateQueryParams();
   }
 
+  protected myEntities(value: boolean) {
```
aamialkovich: I think this is not correct in terms of architecture, as paginated component ideally should know nothing about my entities. I think we need remove it from here and do all logic inside project-list.component

indigo-frontend/src/app/pages/project/project-list/project-list.component.ts
```diff
@@ -106,4 +106,8 @@ export class ProjectListComponent
   onViewChange(view: string) {
     this.selectedView = view as 'grid' | 'list';
   }
+
+  onMyEntitiesOnlyChange(value: boolean) {
+    this.myEntities(value);
```
aamialkovich: I mean something like this

```ts
onMyEntitiesOnlyChange(value: boolean) {
    this.filters['createdByMe'] = value;
    this.dataBh.next([]);
    this.fetchDataAndUpdateQueryParams();
  }
```
Please check if I'm not wrong

aamialkovich (changes requested):

PR #326, opened 2025-10-18
Merge update report service and jrxml template

andrzej16913:
Added more data fields to the new report service and the jrxml template to bring it closer to the old one. Added timezone configurable by editing a value in a config file. Update the report test to reflect the changes. The new report can be found 
[here](https://github.com/user-attachments/files/22984413/report.pdf).

### DIFF
--- a/backend/reports/reports-core/src/main/jasperreports/ExperimentReport.jrxml
+++ b/backend/reports/reports-core/src/main/jasperreports/ExperimentReport.jrxml
@@ -4,8 +4,8 @@
 	<style name="Default Style" default="true" fontName="Arimo" fontSize="10.0"/>
 	<style name="Section Header" style="Default Style"/>
 	<parameter name="logoImage" forPrompting="false" class="java.io.ByteArrayInputStream"/>
-	<parameter name="reportDate" forPrompting="false" class="java.lang.String"/>
-	<parameter name="timeZone" forPrompting="false" class="java.lang.String"/>
+	<parameter name="reportDate" forPrompting="false" class="java.time.ZonedDateTime"/>
+	<parameter name="dateFormat" forPrompting="false" class="java.time.format.DateTimeFormatter"/>
 	<query language="JSON"><![CDATA[]]></query>
 	<field name="project" class="com.epam.indigoeln.eln.model.ProjectDTO"/>
 	<field name="experiment" class="com.epam.indigoeln.eln.model.ExperimentDetailsDTO"/>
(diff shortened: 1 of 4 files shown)


backend/reports/reports-core/src/main/jasperreports/ExperimentReport.jrxml
```diff
@@ -176,7 +177,7 @@
 				</box>
 			</element>
 			<element kind="textField" uuid="3926cfa2-01c8-4d04-be7e-d6cc4519ff2c" stretchType="ContainerHeight" x="141" y="24" width="138" height="20" textAdjust="StretchHeight">
-				<expression><![CDATA[$F{project}.getName()]]></expression>
+				<expression><![CDATA[$F{experiment}.getCreatedAt().withZoneSameInstant(java.time.ZoneId.of($P{timeZone})).format(java.time.format.DateTimeFormatter.ofPattern("MMM d, u HH:mm:ss", Locale.ENGLISH)) + " " + $P{timeZone}]]></expression>
```
denis-dzenskevich: Let's get back to our date formatting.

I tried to write a test to see how timezone is formatted with different time-zone related specifiers of different length:

```java
    @Test
    void testTimeZone() {
        for (String letter : List.of("V", "v", "z", "O", "X", "x", "Z")) {
            for (int count = 1; count <= 4; count++) {
                String spec = letter.repeat(count);
                try {
                    DateTimeFormatter formatter = DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss " + spec).withZone(ZoneId.of("Asia/Istanbul"));
                    String formatted = formatter.format(java.time.Instant.now());
                    System.out.println("Pattern: " + spec + " => " + formatted);
                } catch (Exception e) {
                    System.out.println("Pattern: " + spec + " => Error: " + e.getMessage());
                }
            }
        }
    }
```

Please pick one that suits us better and use it format string. "VV" looks good to me, but please verify what does it show for UTC time zone.

Also, can we use one unified way to format dates? It will simplify adding new dates in the future. Currently report creation date is formatted in Java code and passed as String. Can we pass it as ZonedDateTime and format in the report?

Also, to make things simpler, maybe create DateTimeFormatter in Java code and pass as parameter? To avoid creating it each time in "expression" element. As an added bonus, you can configure DateTimeFormatter to use specific time zone: `DateTimeFormatter.ofPattern(...).withZone(...)`. In this way, it will always use the provided timezone, so you don't have to convert it by hand with `withZoneSameInstant`

denis-dzenskevich (changes requested):

andrzej16913:
I changed the way formatting works. Now DateTimeFormatter is created in ReportService and passed to jrxml template as a parameter. The reportDate is now of type ZonedDateTime and also passed as a parameter.

denis-dzenskevich (approved):

PR #333, opened 2025-10-30
feat(256): [BUG] New notebook is not saved

rmonteseba:
https://github.com/user-attachments/assets/8b5ab67f-42f5-4055-b90f-bc217677f1e7

Added guardrails for ensuring the user meets the name formation criteria + is only able to submit the notebook when the form is valid

image
Message toast for communication in case of error (showing the BE human-readable error)

### DIFF
--- a/indigo-frontend/src/core/components/common/form-dialog/form-dialog.component.html
+++ b/indigo-frontend/src/core/components/common/form-dialog/form-dialog.component.html
@@ -64,6 +64,7 @@
             <eln-button
               [type]="'button'"
               [variant]="'blue'"
+              [disabled]="form.invalid"
               (click)="submit()"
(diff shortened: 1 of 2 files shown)


indigo-frontend/src/core/components/common/form-dialog/form-dialog.component.html
```diff
@@ -64,6 +64,7 @@ <h2 class="text-lg font-semibold text-neutral-900">{{ title }}</h2>
             <eln-button
               [type]="'button'"
               [variant]="'blue'"
+              [disabled]="form.invalid"
```
rmonteseba, line 67: Verified the change across all known usages; observed no functional disruptions on my end

Issue #224, opened 2025-08-03
[In Progress] Add material from reagent catalog


Roman888Kuritsyn:
**Mockups**
[Add material from reagent catalog](https://www.figma.com/proto/RxcXcjpgLOKlORlsWuJ4dR/%F0%9F%8C%9F-Indigo-Designs?node-id=3584-24781&t=giTP83BFFlToDWU7-0&scaling=scale-down-width&content-scaling=fixed&page-id=4040%3A219625&starting-point-node-id=3584%3A24781&show-proto-sidebar=1)
**Description**
Functionality described on the dedicated page: [Add material from reagent catalogue](https://kb.epam.com/display/EPMLSOPELN/Stoichiometry#Stoichiometry-Addmaterialfromreagentcatalogue)

PR #335, opened 2025-10-30
 #224 Add material from reagent catalog

denis-dzenskevich:
- quick search
- structural search
- advanced search
- search results
    - extracted "expandable table" component that allows to expand/collapse multiple rows
    - extracted "infinite loader" not linked to router - maybe it should be integrated into InfiniteLoaderBase, with some flag to disable router-specific functionality

NB: no styling, requires a separate task to make it look like mockups


indigo-frontend/src/app/pages/experiment/experiment-info/experiment-info.component.html
```diff
@@ -79,4 +79,11 @@ <h2 class="font-semibold text-md">Stoichiometric Calculation</h2>
       </eln-card>
     </cdk-accordion-item>
   </cdk-accordion>
+
+  <a
```
aamialkovich: It's better to use button here

indigo-frontend/src/app/pages/experiment/experiment-info/experiment-info.component.ts
```diff
@@ -52,4 +64,18 @@ export class ExperimentInfoComponent implements OnInit {
     // When update completes, refresh the image
     if (!isUpdating) this.experimentImageService.refresh();
   }
+
+  // TODO move to Stoichiometry table when it's available
+  showAddMaterialDialog() {
+    const [e, m] = [this.experiment(), this.model()];
```
aamialkovich: let's replace with `experiment` and `model`

indigo-frontend/src/app/pages/experiment/experiment-info/experiment-info.component.ts
```diff
@@ -52,4 +64,18 @@ export class ExperimentInfoComponent implements OnInit {
...
+    console.log('showAddMaterialDialog', this.experimentModelService, m);
```
aamialkovich: could you please remove?

indigo-frontend/src/app/pages/experiment/sample-search/sample-search.component.ts
```diff
@@ -0,0 +1,327 @@
...
+  @ViewChild('advancedSearchPanel') advancedSearchPanel: MatExpansionPanel;
+
+  service = inject(ApiService);
+  destroyRef = inject(DestroyRef);
+  dialog = inject(MatDialog);
+  experimentModelService = inject(ExperimentModelService);
+
+  title = 'Add Material';
+
+  form = new FormGroup({
```
aamialkovich, line 111: Are there issues with using formly here?

indigo-frontend/src/app/pages/experiment/sample-search/sample-search.component.ts
```diff
@@ -0,0 +1,327 @@
...
+      .request<
+        DictionaryItemRef[]
+      >('get', `dictionaries/${BuiltInDictionary.HEALTH_HAZARD}`)
+      .subscribe((list) => {
+        this.healthHazardsOptions = list;
+      });
+    this.form.valueChanges.pipe(takeUntilDestroyed(this.destroyRef)).subscribe({
+      next: (value) => {
+        console.log('Form changed: ', value);
+        if (value.marked !== this.lastMarked) {
```
aamialkovich: Could you please check if we can use rxjs operator `distinctUntilChanged` before `takeUntilDestroyed` instead of this.

indigo-frontend/src/core/components/common/image/api-image.component.html
```diff
@@ -0,0 +1,9 @@
+@if (src) {
+  <img [src]="src" [alt]="altText" />
```
aamialkovich, line 2: Can we use ngSrc from NgOptimizedImage?

indigo-frontend/src/core/components/common/image/api-image.component.ts
```diff
@@ -0,0 +1,39 @@
...
+import { IsInViewportDirective } from '@core/directives/is-in-viewport.directive';
+import { ApiService } from '@core/services/api.service';
+
+@Component({
+  selector: 'eln-api-image',
+  templateUrl: './api-image.component.html',
+  standalone: true,
+  providers: [
+    {
+      provide: NG_VALUE_ACCESSOR,
```
aamialkovich: Do we need it here?

indigo-frontend/src/core/components/common/numeric-search/numeric-search.component.ts
```diff
@@ -0,0 +1,71 @@
...
+      value: (event.target as HTMLInputElement).valueAsNumber,
+    } as NumericSearch;
+    this.fireOnChange();
+  }
+
+  fireOnChange() {
+    const value = !isNaN(this.value.value)
+      ? { type: this.value.type, value: this.value.value }
+      : null;
+    this.onChange?.(value);
```
aamialkovich: do we need question mark here?

indigo-frontend/src/core/components/experiment/dropdown-value/dropdown-value.component.ts
```diff
@@ -18,13 +19,22 @@ import {
...
+  providers: [
+    {
+      provide: NG_VALUE_ACCESSOR,
+      useExisting: forwardRef(() => DropdownValueComponent),
+      multi: true,
+    },
+  ],
 })
-export class DropdownValueComponent {
+export class DropdownValueComponent implements ControlValueAccessor {
```
aamialkovich, line 30: 👍🏻

indigo-frontend/src/app/pages/experiment/sample-search/sample-search.component.ts
denis-dzenskevich: Yes, works fine, thanks for the suggestion!

indigo-frontend/src/core/components/common/image/api-image.component.html
denis-dzenskevich: Doesn't seem to work with BLOB URLs. I'm using BLOB URL, because I need to load image in JavaScript (otherwise it won't pass the Authentication header). Is there a better solution?

```
error_handler.ts:50 ERROR RuntimeError: NG02952: The NgOptimizedImage directive (activated on an <img> element with the `ngSrc="blob:http://localhost:4200/5b54973c-593a-4bff-9bdd-c44be6ae3e0e"`) has detected that `ngSrc` was set to a blob URL (blob:http://localhost:4200/5b54973c-593a-4bff-9bdd-c44be6ae3e0e). Blob URLs are not supported by the NgOptimizedImage directive. To fix this, disable the NgOptimizedImage directive for this element by removing `ngSrc` and using a regular `src` attribute instead.
```

indigo-frontend/src/core/components/common/numeric-search/numeric-search.component.ts
denis-dzenskevich: Only if this control will be used outside a form, if ever. Removed for now

indigo-frontend/src/core/components/common/image/api-image.component.html
aamialkovich: Got you point, then let's leave as it is 🙂

indigo-frontend/src/app/pages/experiment/sample-search/sample-search.component.ts
denis-dzenskevich: Two issues that gets complicated with Formly:
- using custom controls (ControlValueAccessor)
- custom form layout (labels to the top of the inputs rather then to the left)

PR #342, opened 2025-11-07
#224 Add material from reagent catalog - remarks and small fixes

denis-dzenskevich:


indigo-frontend/src/app/pages/experiment/sample-search/sample-search.component.ts
```diff
@@ -154,12 +155,18 @@ export class SampleSearchComponent implements OnInit {
       .subscribe((list) => {
         this.healthHazardsOptions = list;
       });
+    // when user (de)selects "Only My Materials" when search was already triggered, reload search results
     this.form.valueChanges
-      .pipe(distinctUntilChanged(), takeUntilDestroyed(this.destroyRef))
+      .pipe(
+        map((form) => form.marked),
```
aamialkovich, line 161: Does it mean that we will perform search only on toggler change? Or I'm wrong.

denis-dzenskevich: Search is performed by clicking "Search" button (or pressing Enter). After results are loaded, you'll see "My Materials" checkbox. When you check/uncheck it, it will rerun search, so you don't have to click "Search" again. This block of code is to support this scenario.

(it's not specified in KB, but I thought it will be convenient)

Issue #312, opened 2025-10-08
[Bug] Edit notebook pop-up displayed without previously added values


Roman888Kuritsyn:
<img width="1016" height="460" alt="Image" src="https://github.com/user-attachments/assets/7cc7225e-a711-42c0-8a3b-c0a4657ecde3" />

nadzeyapiatrouskaya-boop:
The bug has still reproduced. Verified Dec 3, 2025. Reopened.

nadzeyapiatrouskaya-boop:
Duplicate: https://github.com/epam/Indigo-ELN-v.-2.0/issues/380

nadzeyapiatrouskaya-boop:
The bug has been fixed successfully. Verified Dec 18, 2025
In confirmation: 

https://github.com/user-attachments/assets/678c4422-feb9-4d74-a2eb-4a2f6cdf1d3a

PR #349, opened 2025-11-19
Ng 312

gunabhushan:
#312 added fix for input fields not being prefilled

### DIFF
--- a/indigo-frontend/src/app/pages/notebook/notebook-edit/notebook-edit.component.ts
+++ b/indigo-frontend/src/app/pages/notebook/notebook-edit/notebook-edit.component.ts
@@ -26,6 +26,7 @@ import { NOTEBOOK_NAME_LENGTH } from '../notebook.constants';
 export class NotebookEditComponent {
   notebookId: string;
   dialogRef = inject(MatDialogRef);
+  model: Partial<Notebook> = {};
   fields: FormlyFieldConfig[] = [
     {
       type: 'input',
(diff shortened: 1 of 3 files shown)


indigo-frontend/src/app/pages/notebook/notebook-edit/notebook-edit.component.ts
```diff
@@ -63,6 +63,12 @@ export class NotebookEditComponent {
     if (data?.notebook?.id) {
       this.notebookId = data.notebook.id;
     }
+    if (data?.notebook?.name) {
```
aamialkovich: I think this is not proper solution. Could you please check `[model]` property from formly? I guess this is what we need.

aamialkovich (changes requested):

indigo-frontend/src/core/components/common/form-dialog/form-dialog.component.ts
```diff
@@ -36,7 +36,7 @@ import { ButtonComponent } from '../button/button.component';
 export class FormDialogComponent {
   dialogRef: MatDialogRef<FormDialogComponent> = inject(MatDialogRef);
   form = new FormGroup({});
-  model: any = {};
+  @Input() model: any = {};
```
aamialkovich, line 39: Can you please replace any type? Also we can rename `model` to `notebook` I guess

PR #400, opened 2025-12-06
I modified the design according to figma

nshan20:


aamialkovich (changes requested):
Please make similar to mockups

indigo-frontend/src/app/pages/experiment/sample-search/sample-search.component.scss
```diff
@@ -0,0 +1,150 @@
+.main-container {
```
aamialkovich: Please use tailwind classes as much as possible.

Issue #353, opened 2025-11-26
[Log In] "Remember Me" checkbox is missing on Log In page


nadzeyapiatrouskaya-boop:
**Description:**
The "Remember Me" checkbox, which allows users to store their credentials for future logins, is not present on the IndigoELN Log In page. According to the [requirements](https://kb.epam.com/display/EPMLSOPELN/Manage+access#Manageaccess-LogIn), this option should be available to users during the login process.

**Steps to Reproduce:**
1.	Navigate to the IndigoELN Log In page: https://indigo-eln-dev.test.lifescience.opensource.epam.com/projects
2.	Observe the available fields and options.

**Actual Result:**
The "Remember Me" checkbox is missing from the Log In page.
<img width="776" height="681" alt="Image" src="https://github.com/user-attachments/assets/ecb2cc96-54e3-406f-87fe-ffd03eb027e1" />

**Expected Result:**
The Log In page should display a "Remember Me" checkbox, allowing users to opt in for storing their credentials:
<img width="616" height="526" alt="Image" src="https://github.com/user-attachments/assets/ca855129-e650-4629-9584-697bd04df497" />
Mockups:
https://www.figma.com/design/RxcXcjpgLOKlORlsWuJ4dR/%F0%9F%8C%9F-Indigo-Designs?node-id=2532-2&p=f&t=5V6yYhh8O5ZewJma-0

Giusha443:
https://github.com/epam/Indigo-ELN-v.-2.0/pull/456

nadzeyapiatrouskaya-boop:
The bug has been fixed partially. Verified 2026-01-26.
As is: "Remember me" checkbox is shown on the Login page, however it is overlapped in situation with incorrect username or password

<img width="807" height="779" alt="Image" src="https://github.com/user-attachments/assets/06cde6b0-f8d4-467c-bd8c-cdc56b67952c" />

Giusha443:
Fixed a problem and waiting for PR review https://github.com/epam/Indigo-ELN-v.-2.0/pull/513

nadzeyapiatrouskaya-boop:
The bug has been fixed successfully. Verified on Dev environment 2026-02-21
In confirmation: 

<img width="1917" height="968" alt="Image" src="https://github.com/user-attachments/assets/66e6095c-79fb-4482-8bb9-9587a1485acf" />

PR #456, opened 2025-12-24
 feat: add checkbox #353

Giusha443:
its working like this (if we check 'remember me' it becomes localstorage data, after closing tab and opening again , it logins automatically and if we don't check 'remember me' it becomes session storage so if we close tab an reopen it will show login page) so I used Amplify`s Cognito, but a problem is a Placement of checkbox , I tried a lot, but could not reach desired position without braking stuff

current situation:
image

CHANGED
its working and according to Figma template
image

### DIFF
--- a/indigo-frontend/src/app/app.component.html
+++ b/indigo-frontend/src/app/app.component.html
@@ -14,6 +14,20 @@
   >
     <router-outlet />
   </ng-template>
+  <ng-template amplifySlot="sign-in-footer">
+    <div class="flex items-center gap-2 px-4 pb-4">
+      <input
+        id="rememberMe"
+        type="checkbox"
+        [checked]="isRemembered"
+        (change)="onRememberMeChange($event)"
+        class="w-4 h-4 text-indigo-600 border-gray-300 rounded cursor-pointer"
+      />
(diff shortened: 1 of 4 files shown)


indigo-frontend/src/app/app.component.html
```diff
@@ -1,4 +1,5 @@
 <amplify-authenticator [hideSignUp]="true">
+  <div class="Body"></div>
```
aamialkovich: could you please use class starting with lower case?

indigo-frontend/src/app/app.component.scss
```diff
@@ -45,3 +46,10 @@
     }
   }
 }
+#rememberme {
+  position: absolute;
+  bottom: 62px;
+  margin-left: 3px;
+  accent-color: var(--primary-500);
+  // margin-bottom: 20px;
```
aamialkovich: Could you please remove unnecessary line?

aamialkovich (approved):

indigo-frontend/src/app/app.component.scss
Giusha443: Done

PR #475, opened 2025-12-29
fix: change placeholder for teams input

Giusha443:
Was: Add users 
Now: Emails, comma separated

image


serhatkaya (approved):
can be merged

Issue #388, opened 2025-12-03
[BUG] There is incorrect option in "More options" modal of the Project card


nadzeyapiatrouskaya-boop:
**Steps to Reproduce:**

1. Open app: https://indigo-eln-dev.test.lifescience.opensource.epam.com/ and login (e.g. as admin)
2. Click "..." button near any project card
3. Observe the action

**Actual result:** 
Modal with " Action" option is opened

<img width="606" height="222" alt="Image" src="https://github.com/user-attachments/assets/b3cc4614-727c-49dc-987a-05c2d640a8d4" />

**Expected result:** 
Modal with "Rename" and "Delete" options is opened

<img width="1242" height="748" alt="Image" src="https://github.com/user-attachments/assets/3068a844-dca5-4023-9557-f632af527598" />

Mockup: https://www.figma.com/design/RxcXcjpgLOKlORlsWuJ4dR/%F0%9F%8C%9F-Indigo-Designs?node-id=22-506&p=f&t=io5lkduRlcoNwirE-0

Giusha443:
https://github.com/epam/Indigo-ELN-v.-2.0/pull/490

nadzeyapiatrouskaya-boop:
The bug has been fixed correctly. Verified on Dev environment 2026-01-31
In confirmation:

<img width="1901" height="921" alt="Image" src="https://github.com/user-attachments/assets/865efc6c-4d77-4a4d-854e-320705ba80ca" />

PR #490, opened 2026-01-07
fix: change action name and styles according to figma #388

Giusha443:
added 2 actions according to figma template


indigo-frontend/src/app/pages/project/project-item/project-item.component.html
```diff
@@ -134,14 +147,20 @@
       >
         <div class="flex -space-x-2">
           @for (member of project.acl.slice(0, 3); track member) {
-            <eln-avatar classList="w-6 h-6" [img]="'https://ui-avatars.com/api/?name=' + (member.username ?? member.displayName)" />
+            <eln-avatar
+              classList="w-6 h-6"
+              [img]="
+                'https://ui-avatars.com/api/?name=' +
```
serhatkaya: You don’t really need a third-party service for this.
A simple pipe that returns the initials from the given string, plus a bit of CSS, is enough to create a clean avatar-like element.

Giusha443: its not my code,i t was given before ,but i will try to fix it as you said

indigo-frontend/src/core/components/common/avatar/avatar.component.ts
```diff
@@ -10,5 +10,6 @@ import { Component, Input } from '@angular/core';
 })
 export class AvatarComponent {
   @Input() img = 'assets/avatar-placeholder.png';
+  // @Input() initials?: string;
```
aamialkovich: Do we need it here?

indigo-frontend/src/app/pages/project/project-item/project-item.component.scss
```diff
@@ -0,0 +1,43 @@
...
+    font-size: 20px !important;
+    margin-right: 8px;
+  }
+  .menu-label {
+    position: absolute;
+    top: 14px;
+  }
+}
+
+.avatar {
```
aamialkovich: Can we use as much as possible tailwind css classes?

Giusha443: Done.

indigo-frontend/src/app/pages/project/project-item/project-item.component.scss
```diff
@@ -0,0 +1,4 @@
+::ng-deep .mat-mdc-menu-panel {
```
aamialkovich: This should be taken carefully as this construct will affect all .mat-mdc-menu-panel classes inside app, so we have two options to make it work just for project-item.component.
1.
```css
:host {
  ::ng-deep {}
}
```

2. add one more class to mat-mdc-menu-panel and specify selector something like

```css
::ng-deep .mat-mdc-menu-panel.some-specific-css-class
```

aamialkovich (approved):
Just small extra comment

PR #491, opened 2026-01-08
[ng-480] Update duplicate error message

smamatov:
1. Updated the error message 
2. Updated the build gradle for core to support Mac


denis-dzenskevich:
Merged via https://github.com/epam/Indigo-ELN-v.-2.0/pull/572

PR #492, opened 2026-01-08
fix: add snak-bar for showing an error with according text

Giusha443:
image
added snakbar,  now validation error message is displayed with the validation rule.

but if we want to use figma`s designed modal we should use:
catchError((editError) => {
          const errorMessage =
            editError.error?.\[0]?.message ||
           editError.error?.message ||
            editError.message ||
            'An error occurred while updating the notebook';
           this.notificationService.notify({
            message: errorMessage,
            type: NotificationType.Error,
            isInline: false,
          });
          return of(null);
        }),
with result:
image

### DIFF
--- a/indigo-frontend/src/app/pages/notebook/notebook-edit/notebook-edit.component.ts
+++ b/indigo-frontend/src/app/pages/notebook/notebook-edit/notebook-edit.component.ts
@@ -11,7 +11,7 @@ import { FormlyFieldConfig } from '@ngx-formly/core';
 import { toHTML } from 'ngx-editor';
 import { catchError, of } from 'rxjs';
 import { NOTEBOOK_NAME_LENGTH } from '../notebook.constants';
-
+import { MatSnackBar } from '@angular/material/snack-bar';
 @Component({
   standalone: true,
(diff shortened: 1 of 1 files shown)


indigo-frontend/src/app/pages/notebook/notebook-edit/notebook-edit.component.ts
```diff
@@ -82,9 +83,10 @@ export class NotebookEditComponent {
       })
       .pipe(
         catchError((editError) => {
-          alert(
-            `Error: ${editError?.error[0].message || 'There was an error updating notebook, please try again later.'}`,
-          );
+          const errorMsg =
```
aamialkovich, line 88: Could you please use notificationService?

aamialkovich (changes requested):

indigo-frontend/src/app/pages/notebook/notebook-edit/notebook-edit.component.ts
Giusha443: Done

aamialkovich (approved):

PR #497, opened 2026-01-12
 MyEntites button fix #444

### LINKED ISSUE #444: [BUG] When the "My Entities" toggle button is switched on, an endless generation of requests occurs: createdByMe=true
**Steps to Reproduce:**

1. Open app: https://indigo-eln-dev.test.lifescience.opensource.epam.com/ and login (e.g. as [alice@eln.com](mailto:alice@eln.com)) with opened DevTools = > Network tab
2. Turn on "My Entities" toggle button
3. Observe the action

**Actual result:** 
When the "My Entities" toggle button is switched on, an endless generation of requests occurs: https://indigo-eln-dev.test.lifescience.opensource.epam.com/api/eln/projects?createdByMe=true&pageSize=10&pageNo=2&sortBy=createdAt&sortOrder=asc

_**Note! When switched back, the projects are displayed on the dashboard in reverse order.**_

https://github.com/user-attachments/assets/cba05dbe-ae0a-47db-bd79-277eddb10335

**Expected result:** 
The IndigoELN System filters the project list to show only those created by the User.
The User views the updated project list.
Requirements:
https://kb.epam.com/display/EPMLSOPELN/Dashboard+and+Navigation#DashboardandNavigation-ViewAllProjects

Image

Giusha443:
now MyEntites button gives expected result, no sorting by latest after clicking it (but there is problem of sorting on website)

CHANGED
sorting is working fine

image

### DIFF
--- a/indigo-frontend/src/app/pages/project/project-list/project-list.component.ts
+++ b/indigo-frontend/src/app/pages/project/project-list/project-list.component.ts
@@ -61,17 +61,17 @@ export class ProjectListComponent
         {
           label: 'Sorting by: Earliest',
           value: 'createdAt',
-          defaultOrder: 'asc',
+          defaultOrder: 'EARLIEST',
         },
         {
           label: 'Sorting by: Latest',
           value: 'createdAt',
-          defaultOrder: 'desc',
+          defaultOrder: 'LATEST',
         },
       ],
       defaultSort: {
         sortBy: 'createdAt',
-        sortOrder: 'asc',
+        sort: 'EARLIEST',
       },
     });
 
@@ -108,7 +108,7 @@ export class ProjectListComponent
   }
 
   onSortChange(event: SortChangeEvent) {
-    this.sort(event.sortBy, event.sortOrder);
+    this.sort(event.sortBy, event.sort);
   }
 
   onViewChange(view: string) {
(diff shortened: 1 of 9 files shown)


indigo-frontend/src/app/pages/project/project-list/project-list.component.ts
```diff
@@ -117,7 +117,7 @@ export class ProjectListComponent
 
   onMyEntitiesOnlyChange(value: boolean) {
     this.filters['createdByMe'] = value;
-    this.dataBh.next([]);
-    this.fetchDataAndUpdateQueryParams();
+    this.pager.pageNo = 0;
```
aamialkovich: Do we also need to do it? Seems it's already done inside `reload` method if I'm not mistaken?

aamialkovich:
@Giusha443 could you please also update asc and desc sorting options in request?

Seems we need to send
```
sort=EARLIEST
sort=LATEST
```
and this should fix the issue

indigo-frontend/src/core/utils/http.util.ts
```diff
@@ -9,11 +9,18 @@ export function getRequestParams(
 
   const filters = { ...filter, ...paging };
 
-  Object.keys(filters).forEach(
-    (key) =>
-      (typeof filters[key] == 'number' || !!filters[key]) &&
-      (params = params.append(key, filters[key])),
-  );
+  Object.keys(filters).forEach((key) => {
```
aamialkovich: What do you think about leaving this method as it was, but update our interfaces and related objects with `EARLIEST` and `LATEST` instead of `asc` and `desc`?

aamialkovich (approved):

PR #500, opened 2026-01-13
Fix/attachments 440 441

LusineNorekyanEDU:
Description

This pull request addresses two issues related to attachment handling in the ELN backend:

Ticket 440: Uploading an attachment to a valid existing project previously resulted in a 500 Internal Server Error.
- Verified that uploads to valid projects now complete successfully and return 200 OK.

Ticket 441: Uploading or deleting attachments for non-existing projects returned a 500 Internal Server Error.
- Fixed by adding proper EntityNotFoundException handling for missing entities.

Changes Made

Added explicit null checks in AttachmentService methods (createProjectAttachment, createNotebookAttachment, etc.) to ensure missing entities trigger a proper 404 response instead of NullPointerException.

Updated unit tests in ProjectServiceTest to cover invalid project scenarios.

Created new test case testUploadAttachmentToInvalidProject verifying that uploading an attachment to a non-existent project returns the correct error response.

Verified all existing attachment-related tests pass for valid projects.

Fixes implemented and validated entirely through unit tests, since the application cannot currently be run locally.

Outcome

Uploads to valid projects now return 200 OK
Uploads or deletions for invalid projects return 404 Not Found
No regressions in existing attachment functionality

### DIFF
--- a/backend/eln/eln-core/src/main/java/com/epam/indigoeln/eln/service/AttachmentService.java
+++ b/backend/eln/eln-core/src/main/java/com/epam/indigoeln/eln/service/AttachmentService.java
@@ -52,6 +52,11 @@ public class AttachmentService {
 
     public List<AttachmentDTO> createProjectAttachment(UUID projectId, String filename, byte[] content) {
         ProjectEntity project = projectRepository.get(projectId);
+
+        if (project == null) {
+            throw new EntityNotFoundException(EntityType.PROJECT, projectId);
+        }
+
         aclService.ensureAccess(project, ApplicationPermission.EDIT_PROJECTS);
         AttachmentEntity attachment = doCreateAttachment(filename, content);
         project.getAttachments().add(attachment);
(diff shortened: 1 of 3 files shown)


backend/eln/eln-core/src/main/java/com/epam/indigoeln/eln/service/AttachmentService.java
```diff
@@ -52,6 +52,11 @@ public List<AttachmentDTO> createProjectAttachment(UUID projectId, FileUpload fi
 
     public List<AttachmentDTO> createProjectAttachment(UUID projectId, String filename, byte[] content) {
         ProjectEntity project = projectRepository.get(projectId);
+
+        if (project == null) {
```
denis-dzenskevich: Can we move these checks to repository classes? It will be more robust as we can easily forget to null-check now or later. Similar to how it's done in loadDetails, it will throw exception (in doLoadDetails) if entity not found

denis-dzenskevich (changes requested):

denis-dzenskevich (approved):

Issue #422, opened 2025-12-12
[BUG] When adding multiple members to a project, the "Add Member" button name changes.


nadzeyapiatrouskaya-boop:
**Steps to Reproduce:**
1. Open app: https://indigo-eln-dev.test.lifescience.opensource.epam.com/ and login (e.g. as [alice@eln.com](mailto:alice@eln.com))
2. Open any project
3. Click dropdown for adding members and add two or more members
5. Take a look at button name

**Actual result:** 
Button name changes to "Add <number> members"
Note! Reproduced on the Notebook level

<img width="501" height="400" alt="Image" src="https://github.com/user-attachments/assets/1cc1a49a-3047-49dd-938a-035a73b8da8f" />

**Expected result:** 
Button name should be "Add Member"

<img width="1351" height="767" alt="Image" src="https://github.com/user-attachments/assets/045bba42-b504-4607-a39b-ef854b206983" />

Mockup:
https://www.figma.com/design/RxcXcjpgLOKlORlsWuJ4dR/%F0%9F%8C%9F-Indigo-Designs?node-id=3417-46807&t=oqA9pDAHoFJaNqHW-0

Giusha443:
https://github.com/epam/Indigo-ELN-v.-2.0/pull/503

nadzeyapiatrouskaya-boop:
The bug has been fixed correctly. Verified on Dev environment 2026-01-31.
In confirmation:

https://github.com/user-attachments/assets/56c1649e-463f-424e-b167-bb638acb7fdb

The missing capital letter in the word "Member" will be fixed as part of a bug: https://github.com/epam/Indigo-ELN-v.-2.0/issues/371

PR #503, opened 2026-01-14
fix: change 'add member' buttons label #422

Giusha443:
now even if it is 1,2,2+ members to add , label is same

image

### DIFF
--- a/indigo-frontend/src/core/components/common/team/team.component.ts
+++ b/indigo-frontend/src/core/components/common/team/team.component.ts
@@ -1,11 +1,28 @@
-import { Component, Input, OnInit, inject, signal, computed, WritableSignal, ViewChild } from '@angular/core';
+import {
+  Component,
+  Input,
+  OnInit,
+  inject,
+  signal,
+  computed,
+  WritableSignal,
+  ViewChild,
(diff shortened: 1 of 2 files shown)


indigo-frontend/src/core/components/common/team/team.component.ts
```diff
@@ -17,174 +34,210 @@ import { TeamComponentConfig } from './team.config';
...
+    }));
+  });
+
+  aclLevelOptions = ELIGIBLE_ACL_LEVELS;
+
+  private api = inject(ApiService);
+
+  @ViewChild(NgSelectComponent) ngSelectComponent!: NgSelectComponent;
+
+  get addMemberLabelMap(): Record<string, string> {
```
aamialkovich: Seems we just can put label inside html and don't use plural pipe

```html
<eln-button>Add member</eln-button>
```

aamialkovich (approved):

indigo-frontend/src/core/components/common/team/team.component.ts
Giusha443: Done.

aamialkovich (approved):

Issue #296, opened 2025-09-25
[BE] Support local run with docker compose


denis-dzenskevich:
- Implement local run mode
    - Backend eln-service
    - Frontend
    - Keycloak for authentication
    - nginx or alternative to serve everything from the same hostname
- Directory "deployment-compose" contains some drafts

PR #519, opened 2026-01-29
#296 local run with docker

m-markau:
This is an interim implemenation.
Nothing is fully working yet.


denis-dzenskevich (changes requested):
Let's merge, to make branches don't diverge too much, I will later add remarks that can be addressed in subsequent PRs

But please make sure frontend is not broken:
- add required packages to package.json and package.json.lock; typically if you do `npm install`, it should be done automatically; right now the build fails to find keycloak-angular
- `npm run start` (Angular development mode) must work with Cognito; if needed, add a separate profile and add new configuration like "start-local" that runs `ng serve` with this new profile

m-markau:
Fixed "npm run start"

denis-dzenskevich (approved):

backend/eln/eln-core/Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM registry.access.redhat.com/ubi8/openjdk-21:1.20
```
denis-dzenskevich, line 1: Can we use eln-service for this? The idea is to have eln-lambda and eln-service with different configuration/integratin options, and eln-core to contain all the business logic

backend/postgres/Dockerfile
```diff
@@ -1,16 +1,19 @@
+#!/bin/bash
```
denis-dzenskevich, line 1: Is it needed in Dockerfile?

deployment-compose/docker-compose.yml
```diff
@@ -1,108 +1,83 @@
...
 
-  eln-service:
+  eln-core:
+    image: "eln-core:development"
     build:
-      context: ./eln/eln-service
-      dockerfile: Dockerfile.jvm
+      context: ./../backend/eln/eln-core
+      dockerfile: Dockerfile
+    command: sh -c "sleep 60 && /opt/jboss/container/java/run/run-java.sh"
```
denis-dzenskevich, line 21: Why we do sleep? If it's to wait for Postgres or Keycloak, can it be done with depends_on?

deployment-compose/postgres/init.sql
```diff
@@ -9,3 +9,7 @@ CREATE DATABASE signature OWNER signature;
 
 CREATE USER eln WITH PASSWORD 'eln';
 CREATE DATABASE eln OWNER eln;
+GRANT CONNECT ON DATABASE eln TO eln;
```
denis-dzenskevich, line 12: eln is already owner, I guess it has all permissions already

indigo-frontend/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+WORKDIR /app
+
+# Copy package.json and package-lock.json to the container
+COPY package.json package-lock.json ./
+
+# Install dependencies
+RUN npm install
+
+# Install Keycloak dependencies
+RUN npm install keycloak-angular@19 keycloak-js
```
denis-dzenskevich, line 12: Can we move this dependencies to package.json? This extra step is not obvious when running Angular development mode locally

indigo-frontend/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+COPY . .
+
+# Install Angular CLI globally to use `ng serve`
+RUN npm install -g @angular/cli
+
+# Expose the default Angular development server port
+EXPOSE 4200
+
+# Start the Angular development server
+CMD ["ng", "serve", "--host", "0.0.0.0", "--port", "4200"]
```
denis-dzenskevich, line 23: Can we use build instead of development mode? `npm run build` to build static files, then use nginx or similar to serve these files

Issue #359, opened 2025-11-27
[BUG] Validation error message does not show correctly when user try to EDIT notebook with already existing name


nadzeyapiatrouskaya-boop:
**Steps to Reproduce:**

1. Open app: https://indigo-eln-dev.test.lifescience.opensource.epam.com/ and login (e.g. as [alice@eln.com](mailto:alice@eln.com))
2. Select any notebook from any project and click "Edit" button
3. Enter already existing notebook name (e.g. "00000004") and click "Submit".
**Actual result:** 
Error message with Failure response is appeared.

<img width="1169" height="793" alt="Image" src="https://github.com/user-attachments/assets/7294c3aa-f48c-4137-89a3-b5a2473ab241" />

**Expected result:** 
Validation error message is displayed with the validation rule.
https://www.figma.com/design/RxcXcjpgLOKlORlsWuJ4dR/%F0%9F%8C%9F-Indigo-Designs?node-id=7796-23732&t=IvHy8G6dCpZMjnmd-1
Example: 

<img width="754" height="667" alt="Image" src="https://github.com/user-attachments/assets/776e70b8-1868-44c6-ae55-0acd16590ca0" />

denis-dzenskevich:
Backend part fixed, now correctly returns 400 with appropriate error message

nadzeyapiatrouskaya-boop:
Verified Dec 16, 2025
As is: 

<img width="1909" height="958" alt="Image" src="https://github.com/user-attachments/assets/f1a1b8f9-02cf-4a6c-ab56-513712c7ddaa" />

Should be:
https://www.figma.com/design/RxcXcjpgLOKlORlsWuJ4dR/%F0%9F%8C%9F-Indigo-Designs?node-id=3193-36682&t=f1iRj7yeOT1EfZce-0

<img width="963" height="540" alt="Image" src="https://github.com/user-attachments/assets/f41b89ec-3d81-4710-84b5-9da1c9ea07c2" />

Giusha443:
https://github.com/epam/Indigo-ELN-v.-2.0/pull/492

nadzeyapiatrouskaya-boop:
The bug was not fixed. Verified 2026-01-31
As is:

<img width="1911" height="934" alt="Image" src="https://github.com/user-attachments/assets/905dc644-3189-41c1-a1df-42a0882061ff" />
[...]

Issue #480, opened 2026-01-03
[BUG] An incorrect error message appears when a Notebook is added with existing name.


nadzeyapiatrouskaya-boop:
**Steps to Reproduce:**

1. Open app: https://indigo-eln-dev.test.lifescience.opensource.epam.com/ and login (e.g. as admin)
2. Select any project 
3. Navigate to Notebooks section
4. Click "Add Notebook" button
5.  Enter existing name (e.g. "12345678") and click "Save" button
6. Observe the action

**Actual result:** 
Toast message with text "Notebook with name '12345678' already exists" is appeared 

<img width="1275" height="801" alt="Image" src="https://github.com/user-attachments/assets/9cd00b89-18d0-4cfa-91f5-4bd0a4c69323" />

**Expected result:** 
Inline error message with text " Unique name is required" is appeared 

Mockups: 
https://www.figma.com/design/RxcXcjpgLOKlORlsWuJ4dR/%F0%9F%8C%9F-Indigo-Designs?node-id=3658-47991&t=qxzij9RInqxIqLyU-0

denis-dzenskevich:
Agreed to have notebook name as validation: backend will implement a method to check name availability, frontend calls this method as async validation

Currently, uniqueness is checked only when creating project/notebook. So frontend doesn't really do any validation, it just displays the error post factum. In the mockups and spec, however, it's described as a validation message in red under the "name" field. So, it should be an async validation on frontend, which requires some method to validate name on backend (e.g. POST /api/eln/notebooks/checkUniqueName or something like this), that frontend can call

NB: we can keep existing backend validation on entity creation, to cover direct API calls or edge cases when name was already taken by another user after frontend performed the validation

denis-dzenskevich:
Backend method implemented

nadzeyapiatrouskaya-boop:
The bug has been fixed successfully
Verified on Dev environment 2026-03-14

<img width="1914" height="905" alt="Image" src="https://github.com/user-attachments/assets/fe7d99d4-c9fd-47b7-a8b3-1395b2deb29d" />

Issue #357, opened 2025-11-27
[BUG] Validation error message does not show correctly when user try to create project with already existing name


nadzeyapiatrouskaya-boop:
**Steps to Reproduce:**

1. Open app: https://indigo-eln-dev.test.lifescience.opensource.epam.com/ and login (e.g. as [alice@eln.com](mailto:alice@eln.com))
2. Click "Add Project"  button and enter already existing name (e.g. "a" ) and click "Submit".

**Actual result:** 
Error message with Failure response 400 is appeared.

<img width="989" height="865" alt="Image" src="https://github.com/user-attachments/assets/5698f610-0495-40e3-9797-3255f0280e9f" />

**Expected result:** 
Validation error message is displayed with the validation rule.
https://www.figma.com/design/RxcXcjpgLOKlORlsWuJ4dR/%F0%9F%8C%9F-Indigo-Designs?node-id=7796-23732&t=IvHy8G6dCpZMjnmd-1
Example: 

<img width="754" height="667" alt="Image" src="https://github.com/user-attachments/assets/776e70b8-1868-44c6-ae55-0acd16590ca0" />

Roman888Kuritsyn:
related to #480

denis-dzenskevich:
Backend method implemented

nadzeyapiatrouskaya-boop:
The bug has been fixed successfully. Verified on Dev environment 2026-03-14

<img width="1901" height="933" alt="Image" src="https://github.com/user-attachments/assets/2393749d-8a1f-4006-a371-378b8c729935" />

PR #527, opened 2026-02-03
Fix: Validation error messages on Notebook and Project pages #359

Giusha443:
Update:
curretnly using methods from BE, Notification working while saving incorrect inputs and serer errors.
added asyncValidator for searching unique name, fields underline working fine.

Project add/edit states:
1)require name 
image
2)Unique name
image
3)256+ length name
image

Notebook add/edit states:
1_require name
image
2) Unique name
image
3)inccorect name
image


aamialkovich (changes requested):
There are several points I would take into consideration, please let me know if it works for you.

1. I would try to use async validators instead of receiving error message after submitting the form, so seems we would even no need service, as all the validation for form controls will be done just in form config. https://formly.dev/docs/examples/validation/unique-value-async-validation/ - this is for validation messages inside form.
2. What do you think about using http interceptor for error handling via toast messages? So we will receive response with error, and interceptor will be responsible for displaying backend error messages?

denis-dzenskevich:
> There are several points I would take into consideration, please let me know if it works for you.
> 
>     1. I would try to use async validators instead of receiving error message after submitting the form, so seems we would even no need service, as all the validation for form controls will be done just in form config. https://formly.dev/docs/examples/validation/unique-value-async-validation/ - this is for validation messages inside form.
> 
>     2. What do you think about using http interceptor for error handling via toast messages? So we will receive response with error, and interceptor will be responsible for displaying backend error messages?

We don't have a backend method to validate name uniqueness, but it's being worked on by Sherzod (#357, #480). Maybe we can do the frontend part here with async validation, but mock the backend call and return always true (or maybe return false for some specific names for testing)

Also, as agreed with BA/design, the error should be displayed near "name" field as other errors, not as a toast notification

@aamialkovich

aamialkovich:
@denis-dzenskevich do you mean that we don't need toast messages here in case we receiving errors? Just controls validation?

denis-dzenskevich:
> @denis-dzenskevich do you mean that we don't need toast messages here in case we receiving errors? Just controls validation?

Yes, for validation before submit (including checking name availability) yes, we use inline messages under fields

There are still cases when submit may fail (e.g. another project with this name was already created after frontend validated the name but before our form was submitted). In this case, it will be a toast notification, in the same way as other backend errors. But it should be very rare in reality

Giusha443:
> > @denis-dzenskevich do you mean that we don't need toast messages here in case we receiving errors? Just controls validation?
> 
> Yes, for validation before submit (including checking name availability) yes, we use inline messages under fields
> 
> There are still cases when submit may fail (e.g. another project with this name was already created after frontend validated the name but before our form was submitted). In this case, it will be a toast notification, in the same way as other backend errors. But it should be very rare in reality

Currently we have only underline text while typing incorrect things not toast mesaage , but we have toast mesaage only after submiting form for name already existing, i get why you had misunderstanding, its not showed in screens

aamialkovich:
Could you remove unnecessary commented code?

indigo-frontend/src/core/components/common/form-dialog/form-dialog.component.ts
```diff
@@ -1,18 +1,22 @@
+// import { NotificationComponent } from '@core/components/common/notification/notification.component';
```
aamialkovich: remove please

aamialkovich (approved):

PR #533, opened 2026-02-06
Adjusted button height and spacing in "Add Project" modal.

nshan20:
Buttons now have consistent size and proper alignment.

### DIFF
--- a/indigo-frontend/src/core/components/common/button/button.variant.ts
+++ b/indigo-frontend/src/core/components/common/button/button.variant.ts
@@ -15,12 +15,13 @@ export const ButtonVariant = cva(
       variant: {
         green: ['bg-green-400', 'hover:bg-green-300'],
         blue: [
-          'bg-primary-400', 
-          'hover:bg-primary-500', 
-          'text-neutral', 
-          'disabled:text-neutral-600', 
-          'disabled:border', 
-          'disabled:border-neutral-300'
+          'bg-primary-400',
+          'hover:bg-primary-500',
+          'text-neutral',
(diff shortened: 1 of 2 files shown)


indigo-frontend/src/core/components/common/button/button.variant.ts
```diff
@@ -36,6 +37,7 @@ export const ButtonVariant = cva(
           'hover:bg-primary-50',
           'focus:bg-neutral-200',
           'disabled:bg-neutral-200 disabled:text-neutral-500',
+          'mr-[12px]',
```
aamialkovich: This is not correct place to add margin. It means that all grey buttons will have margin which is not correct

indigo-frontend/src/core/components/common/button/button.variant.ts
```diff
@@ -15,12 +15,13 @@ export const ButtonVariant = cva(
...
-          'disabled:text-neutral-600', 
-          'disabled:border', 
-          'disabled:border-neutral-300'
+          'bg-primary-400',
+          'hover:bg-primary-500',
+          'text-neutral',
+          'disabled:text-neutral-600',
+          'disabled:border',
+          'disabled:border-neutral-300',
+          'h-[40px]',
```
aamialkovich: Seems this option applicable not only for blue buttons and can be moved to array of properties on 4-11 lines.

aamialkovich (changes requested):
Take a look into comments

aamialkovich (approved):

Issue #430, opened 2025-12-15
[FE] Implement global error handling


aamialkovich:
In scope of this task we need to implement such components https://www.figma.com/design/KjAepepwNqfSv3aGD75S6r/%F0%9F%8C%88-Indigo-UI-KIT?node-id=653-421&p=f&t=nqu8m0kr9dV4r94G-0
First 4 will be appears in the top right corner of the screen as the result of some asynchronous operations
<img width="1073" height="609" alt="Image" src="https://github.com/user-attachments/assets/dc44f604-2b51-43ef-b2c3-0d1098bd7b6e" />
Last 4 can be used as part of any component and placed everywhere.

PR #551, opened 2026-02-17
[FE] Implement global error handling #430

denis-dzenskevich:
- Created global error handler:
  - Takes error message(s) from response JSON; typically, they are more useful and has more details than generic message like "Cannot create project"
  - If JSON is not available (backend malfunctioned or network failed), it shows generic message; ideally we would detect some types of errors like network failures and provide more meaningful errors in this case
- All uncaught RxJS errors are processed by this handler; so in typical case no need to deal with errors at all
- If you need to provide custom handling _in addition to default toast notification_ (e.g. update component state), use `tap({error: ...})`
- If you need to provide custom handling _instead of default notificaiton_, just use standard `pipe(catchError(...))` or `subscribe({error: ...})`. If you don't rethrow error in your handler, it won't be processed by the default handler
- If we want to provide custom error message, use `catchError` and `throwError` with custom error message
- As a downside, we cannot provide default error message per use case (e.g. "Cannot create notebook" or "Cannot update project"); assuming backend will send meaningful error message so user won't get lost. However, if we want to highlight what kind of operation was failed, we likely need to add some pipe operation to every API call. Alternatively, we can add an optional argument to APIService methods

- As an alternative, we could use HTTP interceptor, but from what I understand it's less flexible and wouldn't allow to override error handling behavior per use case

- Unrelated, but I noticed several places where catchError was pushing `null` value downstream; subscribe then was checking for `null` value; I replaced some of these usages, separating next path from error path

### DIFF
--- a/indigo-frontend/src/core/services/notebook/notebook.service.ts
+++ b/indigo-frontend/src/core/services/notebook/notebook.service.ts
@@ -1,6 +1,7 @@
 import { Injectable, signal } from '@angular/core';
 import { NotebookDetail } from '@/core/types/entities/notebook-detail.i';
 import { ApiService } from '@/core/services/api.service';
+import { finalize, tap } from 'rxjs';
 
 @Injectable()
 export class NotebookService {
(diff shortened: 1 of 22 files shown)


indigo-frontend/src/core/services/notebook/notebook.service.ts
```diff
@@ -13,25 +14,32 @@ export class NotebookService {
...
     this.isLoading.set(true);
     this.hasError.set(false);
 
-    this.api.request<NotebookDetail>('get', `notebooks/${id}`).subscribe({
-      next: (nb) => {
+    this.api
+      .request<NotebookDetail>('get', `notebooks/${id}`)
+      .pipe(
+        tap({
+          error: () => this.isLoading.set(false),
```
aamialkovich: Do we need it here? Seems we already do it in finalize

denis-dzenskevich: My bad, moved the wrong expression to "error"

PR #552, opened 2026-02-18
Update the Signature stamping

andrzej16913:
Changed the reject functionality to align more closely to how signing is implemented, so it will leave a visible stamp on a document.


backend/signature/signature-core/src/main/java/com/epam/indigoeln/signature/service/signatureapplier/AbstractSigner.java
```diff
@@ -1,59 +1,83 @@
...
+        appearance.setReason(reason);
+        appearance.setSignDate(signDate);
+    }
+
+    public void closeSignatureAppearance(PdfSignatureAppearance appearance) throws IOException {
+        Map<PdfName, Integer> exc = new HashMap<>();
+        exc.put(PdfName.CONTENTS, 10);
+        appearance.preClose(exc);
+        PdfDictionary update = new PdfDictionary();
+        update.put(PdfName.CONTENTS, new PdfString("aaaa").setHexWriting(true));
```
denis-dzenskevich: What is "aaaa"?

backend/signature/signature-core/build.gradle.kts
```diff
@@ -23,7 +23,7 @@ dependencies {
     implementation("org.bouncycastle:bcpkix-jdk18on")
 
     implementation("one.util:streamex:0.8.3")
-    implementation("com.itextpdf:itextpdf:5.5.13.4")
+    implementation("com.github.librepdf:openpdf:3.0.1")
```
denis-dzenskevich: Can we use Quarkus OpenPDF extensions ("io.quarkus:...")?

denis-dzenskevich (changes requested):
I tried opening the PDF in Acrobat Reader, and here's what is shows me ("Signatures" panel):

Image

1. Why the text is "Signature Not Verified"?
2. Why reject signature is on the second page? The file was only 1 page long, maybe we can place several signatures on one page? Also, ideally it won't obscure the text, but it depends on where the text is, so we can reposition it later to look better on our experiment reports
3. There are 3 errors for each signature. The one about Signer's identity is expected, because we use self-signed certificate to sign. But what about two other errors? What can be the cause and how to fix it? Maybe the old implementation had the same errors, I didn't try

For reference, here's the text of the first error: Signature contains incorrect, unrecognized, corrupted or suspicious data. Support Information: SigDict /SubFilter value

denis-dzenskevich (approved):

PR #568, opened 2026-02-25
Add notebook name existence check #480

MeriGasparyan:


backend/eln/eln-core/src/main/java/com/epam/indigoeln/eln/repository/NotebookRepository.java
```diff
@@ -70,6 +70,12 @@ public boolean hasAccessibleNotebooks(ProjectEntity project) {
         return find("project", project).firstResult() != null;
     }
 
+    public boolean existsByName(String name) {
```
denis-dzenskevich, line 73: Probably not very important here, as it likely won't be called very often, but can you please replace with "exists" check? So we don't load entire entity from the DB where we only need to know exists or not?

denis-dzenskevich (changes requested):

backend/eln/eln-core/src/main/java/com/epam/indigoeln/eln/repository/NotebookRepository.java
MeriGasparyan: Changed to use count() instead

denis-dzenskevich (approved):

PR #569, opened 2026-02-25
Add project name existence check #357

MeriGasparyan:


backend/eln/eln-core/src/main/java/com/epam/indigoeln/eln/repository/ProjectRepository.java
```diff
@@ -66,6 +66,12 @@ public TotalCounts getTotalCounts() {
         return projectMapper.convertTotalCounts(entity);
     }
 
+    public boolean existsByName(String name){
+        return find("name", name)
```
denis-dzenskevich: Probably not very important here, as it likely won't be called very often, but can you please replace with "exists" check? So we don't load entire entity from the DB where we only need to know exists or not?

denis-dzenskevich (changes requested):

backend/eln/eln-core/src/main/java/com/epam/indigoeln/eln/repository/ProjectRepository.java
MeriGasparyan: Changed to use count instead

denis-dzenskevich (approved):

Issue #394, opened 2025-12-04
[BUG][BE] User without Administrators role can create new user


nadzeyapiatrouskaya-boop:
1. Authorize as user w/o Admin permissions (e.g. as alice@eln.com)
2.  Create POST request {{var}}api/eln/users 
4. Take a look at the response

**Actual result:** 
Status code is 200 OK. User is created.

**Expected result:**
Status code is 403 Forbidden. User is not created.

nadzeyapiatrouskaya-boop:
The bug has been fixed successfully. Verified on Dev environment 2026-03-14.

> [
    {
        "message": "Operation not permitted: MANAGE_USERS"
    }
]

<img width="1104" height="677" alt="Image" src="https://github.com/user-attachments/assets/cf655801-58d7-456d-a340-d1fcd75ced82" />

PR #584, opened 2026-03-10
Fix bug: User without Administrators role can create new user

andrzej16913:
Link to issue: https://github.com/epam/Indigo-ELN-v.-2.0/issues/394

### DIFF
--- a/backend/eln/eln-core/src/main/java/com/epam/indigoeln/eln/service/UserService.java
+++ b/backend/eln/eln-core/src/main/java/com/epam/indigoeln/eln/service/UserService.java
@@ -92,6 +92,10 @@ public class UserService {
     }
 
     public UserDTO createUser(UserRequest request) {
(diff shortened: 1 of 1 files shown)


backend/eln/eln-core/src/main/java/com/epam/indigoeln/eln/service/UserService.java
```diff
@@ -92,6 +92,10 @@ public List<UserRef> suggestUsers(@Nullable String search) {
     }
 
     public UserDTO createUser(UserRequest request) {
+        UserInfo currentUser = getCurrentUser();
```
denis-dzenskevich: Can you please use aclService.ensureTopLevelAccess, to make it uniform with the other places?

denis-dzenskevich (changes requested):

denis-dzenskevich (approved):

Issue #557, opened 2026-02-22
[BUG] An incorrect error message appears when a Notebook is added with an invalid name.


nadzeyapiatrouskaya-boop:
**Description:**

Error in Console occurs when editing a Notebook and leaving the Description field empty.

**Steps to Reproduce:**

1. Open app: https://indigo-eln-dev.test.lifescience.opensource.epam.com/ and login (e.g. as admin) with opened DevTools (Console tab)
2. Select any project 
3. Navigate to Notebooks section
4. Select any Notebook with empty Description
5. Click "Edit" button
6. Do not change anything, just click "Submit" button
7. Observe the action
**Actual result:** 
An unhandled frontend error occurs, and the notebook is not saved. The following error appears in the console:

> ERROR RangeError: Invalid input for  chunk-M5RMYJYT.js:7 
    Node.fromJSON
    at i.fromJSON (chunk-I4M6NR36.js:1:19901)
    at In.nodeFromJSON (chunk-I4M6NR36.js:2:9368)
    at Up (chunk-I4M6NR36.js:10:36873)
    at i.editNotebook (chunk-ADLGUC7K.js:1:1869)
    at chunk-ADLGUC7K.js:1:2466
    at Dl (chunk-M5RMYJYT.js:7:104456)
    at Object.o [as next] (chunk-M5RMYJYT.js:7:104626)
    at Di.next (chunk-M5RMYJYT.js:3:3101)
    at Be._next (chunk-M5RMYJYT.js:3:2784)
    at Be.next (chunk-M5RMYJYT.js:3:2511)

> 

<img width="1908" height="838" alt="Image" src="https://github.com/user-attachments/assets/02fdce61-9d5c-47b7-b0eb-9e641fec77b7" />

https://github.com/user-attachments/assets/c273dd48-c962-4f69-a368-9665f6960fb2

**Expected result:** 
Modal is closed. There are not any errors in the console tab

nadzeyapiatrouskaya-boop:
The bug has been fixed successfully. Verified on Dev environment 2026-03-22.
In confirmation:

<img width="1894" height="1022" alt="Image" src="https://github.com/user-attachments/assets/0c8fbf41-8d9f-4bc5-8c1f-f21667342a03" />

Issue #361, opened 2025-11-27
[BUG] User does not redirect to Project Info page after creating new project


nadzeyapiatrouskaya-boop:
**Description:**

After successfully creating a new project in the IndigoELN system, the user is not redirected to the newly created Project Info page as specified in the requirements. The success message ("Project successfully created.") is also not displayed. Instead, the user remains on the "All projects" dashboard widget, without any notification about the result of the operation.

**Steps to Reproduce:**

1. Open app: https://indigo-eln-dev.test.lifescience.opensource.epam.com/ and login (e.g. as [alice@eln.com](mailto:alice@eln.com))
2. Click "Add Project"  button, fill required fields with correct data and click "Submit".

**Actual result:** 
Popup window is closed. Project is saved and added on the projects dashboard. User stays on the "All projects" dashboard widget.
Success message: "Project successfully created." is not displayed.

https://github.com/user-attachments/assets/5f113e80-e01d-4f32-a23a-640806c61548

**Expected result:** 
If validation passes, the IndigoELN System:
Saves the new Project.
Redirects the User to the newly created Project Info page.
Displays a success message: "Project successfully created."

 **AC N7 of requirements**: https://kb.epam.com/display/EPMLSOPELN/Manage+Projects#ManageProjects-CreateProject

<img width="1038" height="501" alt="Image" src="https://github.com/user-attachments/assets/add5f4d3-afd7-4982-b1b5-65634e8a6203" />

nadzeyapiatrouskaya-boop:
The bug has been fixed partially. Verified Dec 16, 2025
As is:

https://github.com/user-attachments/assets/03799326-964c-438b-94a2-8fdaad6cae98
 Should be:
Toast message should look like 
<img width="578" height="119" alt="Image" src="https://github.com/user-attachments/assets/5d76c641-5398-4114-9532-c3158e78ea94" />
Indigo UI KIT: 
https://www.figma.com/design/KjAepepwNqfSv3aGD75S6r/%F0%9F%8C%88-Indigo-UI-KIT?node-id=653-421&p=f&t=1t8Az819ZASnwgrt-0

Giusha443:
https://github.com/epam/Indigo-ELN-v.-2.0/pull/588   added toast message
and also fixed this issue in it https://github.com/epam/Indigo-ELN-v.-2.0/issues/557

nadzeyapiatrouskaya-boop:
[...]

PR #588, opened 2026-03-11
fix: add success notification and refactor unique name with signal

Giusha443:
image
image
image
 now for unique name we use signal for better change detection and to avoid a runtime error in console. and created new service for success notification.


Giusha443:
made  issues https://github.com/epam/Indigo-ELN-v.-2.0/issues/557 and https://github.com/epam/Indigo-ELN-v.-2.0/issues/361

indigo-frontend/src/core/services/notification/success.notification.component.ts
aamialkovich, line 1: I would remove this service as it looks like overengineering.
I would just leave such calling in all places
```
this.notificationService.notify({
  message: 'Write necessary message here',
  type: NotificationType.Success,
  isInline: false,
});```

aamialkovich (changes requested):

indigo-frontend/src/core/services/notification/success.notification.component.ts
Giusha443: made as u asked.

aamialkovich (approved):

PR #595, opened 2026-03-17
feat(570): [FE] Update mutations to use strings instead of numbers

rmonteseba:
image

Number are now handled and sent as strings

### DIFF
--- a/indigo-frontend/src/app/pages/experiment/stoichiometry/product-batch-summary-table/product-batch-summary-table.component.ts
+++ b/indigo-frontend/src/app/pages/experiment/stoichiometry/product-batch-summary-table/product-batch-summary-table.component.ts
@@ -1,5 +1,4 @@
 import { Component, computed, inject, input } from '@angular/core';
-import { Observable } from 'rxjs';
 import {
   ExperimentModel,
   Reaction,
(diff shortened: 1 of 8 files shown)


indigo-frontend/src/app/pages/experiment/stoichiometry/product-batch-summary-table/product-batch-summary-table.component.ts
```diff
@@ -244,31 +242,6 @@ export class ProductBatchSummaryTableComponent {
     return statusMap[status] ?? status;
   }
 
-  private applyUnitInputChange(
```
rmonteseba, line 247: Dead code

denis-dzenskevich (approved):

denis-dzenskevich (changes requested):
I tried to run the code, updated input weight in stoichiometry table, and it seems like it's still being sent as a number (I tried it couple of times, so it's possible that it only affects the second and subsequent updates?)

Image

denis-dzenskevich:
Can you please also update EnteredValue interface to have value: string

denis-dzenskevich (approved):

PR #577, opened 2026-03-04
Fix values accesses for product batch detail

rmonteseba:
Fix values accesses for product batch detail

PR #599, opened 2026-03-20
Rework frontend mutations frontend and add highlighting of updated values

denis-dzenskevich:
- centralize experiment loading and mutation handling in ExperimentDetailsService:
    - make it a singleton to make sure entire application works with a single copy of data
    - make it responsible for restoring previous model in case of errors
    - FUTURE: will be able to handle multiple mutations if user quickly enters several numbers (e.g. will be able to queue mutation and execute one by one)

- added ability to define list of additional classes to the cells of EditableTableComponent; used to highlight updated and overwritten values

- move loading of experiments from components to ExperimentLayout (will be required later: we will load template and then create components and tabs based on template)

- added "Loading" display to EditableTableComponent to prevent showing misleading "No data" while experiment is loading; shall be mostly useless because of the previous one, but I'd leave it (e.g. if this component will be used in other places)

NB: merge after https://github.com/epam/Indigo-ELN-v.-2.0/pull/577


indigo-frontend/src/core/services/experiment/experiment-detail.service.ts
```diff
@@ -106,6 +83,7 @@ export class ExperimentDetailService {
   reset() {
```
rmonteseba, line 83: Maybe add `this.isUpdating.set(false);` here?

indigo-frontend/src/core/services/experiment/experiment-detail.service.ts
```diff
@@ -31,14 +34,14 @@ export class ExperimentDetailService {
       .subscribe({
         next: (exp) => {
           this.experimentDetail.set(exp);
```
rmonteseba, line 36: `lastLoadedDetail.set(updated)` stores the same reference as `experimentDetail`, so if updated is ever mutated in place, both signals would be affected and the revert on error could end up restoring already-mutated data. Could be worth using `structuredClone(updated)` here, same as you already do in `load()`, just to be safe.

rmonteseba (approved):

indigo-frontend/src/core/services/experiment/experiment-detail.service.ts
denis-dzenskevich: Done!

indigo-frontend/src/core/services/experiment/experiment-detail.service.ts
denis-dzenskevich: Makes sense, thanks for noticing!

Issue #222, opened 2025-08-03
[In Progress] Analyze RXN


Roman888Kuritsyn:
**Mockups**
[Analyze RXN](https://www.figma.com/proto/RxcXcjpgLOKlORlsWuJ4dR/%F0%9F%8C%9F-Indigo-Designs?node-id=5980-67886&t=lsf1UOxRAdAnCrQm-0&scaling=scale-down-width&content-scaling=fixed&page-id=5980%3A50142&starting-point-node-id=5980%3A67865&show-proto-sidebar=1)
**Description**
Functionality described on the dedicated page: [Analyze RXN](https://kb.epam.com/display/EPMLSOPELN/Reaction+schema#Reactionschema-AnalyseRxn)

Roman888Kuritsyn:
#224
related

Issue #598, opened 2026-03-19
[Ready] Retrieve materials from Pubchem API


Roman888Kuritsyn:
Search option on Add material screen should request data from Pubchem API 
PubChem search integration should allow users to search for substances from ELN by Structure in SMILES format (entered via Draw Stucture), 
CAS or name (entered in text input).
[PubChem](https://kb.epam.com/spaces/EPMLSOPELN/pages/2790843650/PubChem+search)

Roman888Kuritsyn:
Search for Indigo 2.0 was implemented in #24 and #22

Issue #550, opened 2026-02-16
[FE] Merge Reaction Schema and Reaction Editor into single component


Roman888Kuritsyn:

denis-dzenskevich:
Please use eln-api-image component to display reaction, with the following URL:

`/api/eln/experiments/{experimentId}/datamodel/reactions/{reactionAnchor}/picture?version={rxnVersion}`

Where rxnVersion is taken from reaction. It will automatically update image when reaction is changed

PR #612, opened 2026-04-06
#222 Analyze RXN

denis-dzenskevich:
Lot of changes, but it was somewhat hard to test the feature without it. Plus some preparations for upcoming changes (build experiment screen according to the components in template, PubChem and support of multiple catalogs in Search Sample Dialog)

- updated API to return MutationResponse with some additional fields:
  - for SetScheme mutation, return inputs without samples with their structure to be used for subsequent search
  - add messages to be displayed for successful mutation (can be useful to clarify something to the user)

- refactored slide-in panel into a separate component, so it can be used like Angular dialog; currently used for Analyze RXN and Add Material dialogs, but our UX likes sliding panels, so will be useful in other places too
  - it's slightly complex; maybe someone has suggestions how to simplify it?

- fixed Sample Search dialog that got broken at some point
  - extracted search results into a separate component to reuse it in Analyze RXN

- we discussed future designs with BAs, and it will affect Stoichiometry table: first, it will be two-level (upper level for compounds and lower level for samples), and it will use per-row buttons to delete items (instead of checkboxes plus table-wide button); our checkbox ticket got stuck somewhere, so I switched the table to use delete buttons

- slightly simplified how SetSchema mutation was created

- added spinner to infinite loaders; it was too scary to see all your experiments are gone, until you realize they were just loading without any indication


indigo-frontend/src/app/pages/experiment/sample-search-results/sample-search-results.component.html
```diff
@@ -0,0 +1,109 @@
...
+            bookmark
+          </mat-icon>
+        }
+      </ng-template>
+      <ng-template elnColumnDef="add" displayName="" let-element>
+        <button
+          class="ml-2 px-4 py-2 bg-[#f3f3f3] text-white rounded-[5px] cursor-pointer"
+        >
+          <mat-icon
+            (click)="addToExperiment.emit(element)"
```
rmonteseba: Can this `(click)` be in the button for more consistency?

indigo-frontend/src/app/pages/experiment/stoichiometry/reaction-scheme-view/reaction-scheme-view.component.ts
```diff
@@ -50,46 +57,29 @@ export class ReactionSchemeViewComponent {
...
-        console.error('Failed to update experiment model:', error);
-        // Notify parent that update completed (with error)
-        this.modelUpdating.emit(false);
-      },
-    });
+  private updateExperiment(mutation: Mutation): void {
+    this.experimentDetailService
+      .updateDataModel(mutation)
+      .subscribe((response) => {
+        this.slideInPanelService.open(AnalyzeRxnComponent, {
```
rmonteseba: This opens Analyze RXN unconditionally, but `unresolvedInputs` is optional. In case the mutation succeeds and the backend returns no unresolved inputs, the panel can open with no actionable data and `selectedTab` may end up as `-1`. Do you think would be ok to guard this flow and only open the panel when `unresolvedInputs` is present and non-empty?.

indigo-frontend/src/app/pages/experiment/stoichiometry/reaction-view/reaction-view.component.html
```diff
@@ -3,7 +3,6 @@
     <eln-reaction-scheme-view
       [reaction]="reaction()"
       [experimentId]="experimentId()"
-      (modelUpdating)="onModelUpdating($event)"
```
rmonteseba, line 6: Could this leave the experiment image stale after editing the reaction? `ExperimentInfoComponent` still has `onModelUpdating()` calling `experimentImageService.refresh()`, but I no longer see that event wired up here. Would it be worth either reconnecting the event or refreshing the image right after a successful `SetScheme` so this lives in one place?

indigo-frontend/src/app/pages/experiment/sample-search/sample-search.component.ts
```diff
@@ -165,13 +170,13 @@ export class SampleSearchComponent implements OnInit {
...
         this.formNotEmpty = Object.entries(formValues)
-          .filter(([k, _]) => k !== 'structureSearchType' && k !== 'marked')
+          .filter(([k, _]) => k !== 'structureSearchType')
           .some(([_, v]) => isFormValueNotEmpty(v));
       });
 
     this.form.valueChanges
       .pipe(
-        map((form) => form.marked),
+        map((form) => form.catalog),
```
rmonteseba, line 179: Is this meant to be future-facing for now? Changing `catalog` retriggers the search, but I do not see that value affecting the request or switching the data source in `performSearch()`, and the same catalog radios in Analyze RXN also look UI-only. If this is not wired end-to-end yet, would it be safer to hide or disable the control so the UI does not imply behavior that is not there yet? Just curious and totally up to you!

rmonteseba:
Showing a spinner here makes the loading state much clearer and avoids the “everything disappeared” feeling while more results are still being fetched. This is a small change, but it adds real UX value. Good stuff here 👍🏼

indigo-frontend/src/app/pages/experiment/sample-search/sample-search.component.ts
denis-dzenskevich: Yes, we don't have backend support for multiple catalogs yet, it will be implemented with PubChem search (#598)

But I disabled PubChem and My Materials radio buttons, so user is not able to select them yet

indigo-frontend/src/app/pages/experiment/stoichiometry/reaction-scheme-view/reaction-scheme-view.component.ts
denis-dzenskevich: Make sense, added a guard

indigo-frontend/src/app/pages/experiment/stoichiometry/reaction-view/reaction-view.component.html
denis-dzenskevich: I removed onModelUpdating fired from a component because updating should be handled by signals in experiment-details service

But also reaction image update is being handled in #550, it will be done in ReactionViewComponent. 

ExperimentInfoComponent shouldn't handle the image, I removed all leftovers not removed previously (onModelUpdating and signals)

rmonteseba (approved):

PR #614, opened 2026-04-07
296: keycloak configs

m-markau:
Keycloak configs for docker installation


deployment-compose/docker-compose.yml
```diff
@@ -71,31 +64,52 @@ services:
...
-    healthcheck:
-      test: ["CMD", "./healthcheck.sh"]
-      start_period: 120s
-      start_interval: 1s
     ports:
       - "8088:8080"
-      #- "8443:8443"
+      - "9005:9005"
+
+  keycloak-healthcheck:
```
denis-dzenskevich, line 87: Do we really need a whole new container just for healthchecks? If we have /health/started endpoint, can't we just use wget/curl as healthcheck for keycloak container?

m-markau: Keycloak container doesn't have any tools installed like wget or curl due to secuity reasons. Thus this is not possible to check health from withing container itself. External healcheck container is common practice for Keycloak

denis-dzenskevich (approved):

PR #616, opened 2026-04-09
Set ExperimentSignificantFigures added

maniepam47:
https://github.com/orgs/epam/projects/41/views/1?filterQuery=significant&pane=issue&itemId=171413223&issue=epam%7CIndigo-ELN-v.-2.0%7C609


indigo-frontend/src/app/pages/experiment/stoichiometry/editable-data-table/editable-data-table.component.html
```diff
@@ -1,7 +1,13 @@
 <div class="flex justify-between py-1">
   <h3 class="text-sm font-semibold p-3">{{ title() }}</h3>
   @if (showAddButton()) {
-  <div class="flex justify-between items-center">
+  <div class="flex justify-between items-center gap-3">
```
denis-dzenskevich: When you open an experiment, significant figures is blank. Please make it display the current value. It's stored in experiment-details service, in `experimentModel().significantFigures`. 

Please also add the corresponding field in interface ExperimentModel

indigo-frontend/src/app/pages/experiment/stoichiometry/editable-data-table/editable-data-table.component.ts
```diff
@@ -120,5 +134,40 @@ export class EditableDataTableComponent<TRow = unknown> {
...
+onSignificantFiguresChange(value: string | string[] | null): void {
+  const parsedValue = this.parseSignificantFigure(value);
+  if (parsedValue === null) return;
+
+  this.experimentDetailService
+    .updateDataModel({
+      type: 'SetExperimentSignificantFigures',
+      significantFigures: parsedValue,
+    })
+    .pipe(
```
denis-dzenskevich: Please remove messages. In case of errors, we have a common handler for all backend errors that will show a notification, and for success case, user will see a lot of numbers updated, so I think it the message will be excessive

denis-dzenskevich (changes requested):

denis-dzenskevich:
PLease rename the PR. Why this strange name?

indigo-frontend/src/app/pages/experiment/stoichiometry/editable-data-table/editable-data-table.component.html
maniepam47: done

indigo-frontend/src/app/pages/experiment/stoichiometry/editable-data-table/editable-data-table.component.ts
maniepam47: done

indigo-frontend/src/core/types/entities/experiments/mutation.i.ts
```diff
@@ -312,6 +312,11 @@ interface SetOutputStructureComment extends ReactionOutputSampleMutation {
   structureComment: string | null;
 }
 
+interface SetExperimentSignificantFigures extends BaseMutation {
+  type: 'SetExperimentSignificantFigures';
+  significantFigures: number | null;
```
denis-dzenskevich: significantFigures cannot be set to null (see ExperimentMutation.java:33)

indigo-frontend/src/core/types/entities/experiments/experiment.i.ts
```diff
@@ -114,4 +114,5 @@ export interface Reaction {
 // ================================
 export interface ExperimentModel {
   reactions: Reaction[];
+  significantFigures?: number;
```
denis-dzenskevich: significantFigures is always present (5 by default), you can remove "?"

indigo-frontend/src/app/pages/experiment/stoichiometry/editable-data-table/editable-data-table.component.ts
```diff
@@ -55,9 +69,14 @@ import {
     FormsModule,
     CommonModule,
     ButtonComponent,
+    SelectComponent,
   ],
 })
 export class EditableDataTableComponent<TRow = unknown> {
+  experimentDetailService = inject(ExperimentDetailService);
+  readonly experimentModel = this.experimentDetailService.experimentModel;
+  readonly selectedSignificantFigure = signal<string | null>(null);
```
denis-dzenskevich: What if we don't use a special signal for selectedSignificantFigure and bind to experimentModel().significantFigures directly?

Actually, even if update experimentModel().significantFigures in place (not sure if our select component does it or no), ExperimentDetailService.updateDataModel will automatically revert to last loaded model in case of errors

What do you think?

denis-dzenskevich (changes requested):

indigo-frontend/src/core/types/entities/experiments/mutation.i.ts
maniepam47: done

indigo-frontend/src/app/pages/experiment/stoichiometry/editable-data-table/editable-data-table.component.ts
maniepam47: yes,we can do either way but in latest angular version we have this feature where any UI ops can be updated like this
i will keep like normal signal only

indigo-frontend/src/app/pages/experiment/stoichiometry/editable-data-table/editable-data-table.component.html
```diff
@@ -1,7 +1,14 @@
...
+  Significant Figures : <eln-select
```
denis-dzenskevich: I just noticed, you added it to editable-table component, which is the base for several tables. In the spec, it says significant figures should be a part of Stoichiometry table only. Can you please move to appropriate component?

denis-dzenskevich (changes requested):

denis-dzenskevich (approved):

PR #617, opened 2026-04-10
296: frontend auth via keycloak

m-markau:
Authentication via Keycloak for local installation of Indigo Frontend


indigo-frontend/src/app/app.routes.local.ts
```diff
@@ -5,7 +5,7 @@ export const routes: Routes = [
   {
     path: '',
     loadComponent: () =>
-      import('@core/components/layout/master.component').then(
+      import('@core/components/layout/master-keycloak.component').then(
```
denis-dzenskevich: I would postpone it until we have some minimal working version, but I think it's getting harder than needed to keep all these files in sync. If you compare app.routes.ts with app.routes.local.ts, where are already couple of differences that didn't get caught

Can we instead substitute master.component.ts with master.component.local.ts?

Or even better, if you compare master.component, it's almost identical, it just calls KeycloakService methods instead of AuthenticatorService. Can we try to combine such differences into one file e.g. user.service.ts vs user-keycloak.service.ts (or identity.service.ts vs identity-keycloak.service.ts) so we don't have to keep (and update) a mirror versions of all these files?

indigo-frontend/src/core/components/layout/master-keycloak.component.html
```diff
@@ -0,0 +1,50 @@
+<div class="flex flex-col h-screen">
```
denis-dzenskevich: No differences with non-Keycloak version?

indigo-frontend/package.json
```diff
@@ -3,7 +3,8 @@
   "version": "0.0.0",
   "scripts": {
     "ng": "ng",
-    "start": "ng serve",
+    "start": "ng serve --configuration production",
```
denis-dzenskevich: I guess it should be development configuration? Otherwise it ruins Angular dev mode

denis-dzenskevich (changes requested):

indigo-frontend/src/app/app.routes.local.ts
denis-dzenskevich: Also, I just thought. Would it simplify the development if we keep both AWS and Keycloak versions in the sources for now, without files substitutions? Based on authProvider, you will call one of the implementations. And when we have working version, we will take a look at the diff and see what really should be changed

m-markau: I have removed duplicated master-keycloak.component and app.routes.local.ts files
Work fines without them. They were needed before swithiching to IdentityService

indigo-frontend/package.json
m-markau: Swithed to "ng serve --configuration development"4
NB! This config doesn't support auth via Cognito to AWS

denis-dzenskevich: "development" seems already the default, so no need to specify it (see angular.json, "defaultConfiguration" for "serve" operation)

But it's broken. Please make sure frontend developers can still use `npm run start` to run development configuration

image

m-markau: Fixed. Please be aware that "npm run start" command by default starts "Production" configuration, not "Development".

indigo-frontend/src/core/components/layout/master-keycloak.component.html
m-markau: Removed file "master-keycloak.component.html" as duplicate. Working fine without it

denis-dzenskevich (approved):

Issue #615, opened 2026-04-09
[BUG] Infinite scroll doesn't load next pages


denis-dzenskevich:
**Scenario**
- open new tab in browser
- open main page of the application (projects list) https://indigo-eln-dev.test.lifescience.opensource.epam.com/
- scroll to the bottom of the page

**Expected**
- all projects (22 for now) are shown

**Actual**
- sometimes only 10 or 20 projects are shown

**Analysis**
The application uses "infinite loader", e.g. it loads the first page, and when user scrolls to the end of page, it transparently load the next page, so it looks like a long list without pagination, despite pagination is used internally.

The application tries to guess number of pages to load initially. Presumably it loads 10 or 20 items, and instead of loading next page, it stops for some reason

It's rarely reproduced, likely only when opening a new tab in browser, maybe it only happens with some window sizes or zoom levels

PR #629, opened 2026-04-14
fix(#615): fix infinite scroll not loading next pages

maravagyan:


indigo-frontend/src/app/pages/project/project-list/project-list.component.html
```diff
@@ -16,7 +16,7 @@
     'md:grid-cols-2 2xl:grid-cols-3': selectedView === 'grid',
   }"
 >
-  @for (project of data; track project.id) {
+   @for (project of (data$ | async) ?? []; track project.id) {
```
denis-dzenskevich, line 18: Shall we do the same in other list views (you can search when eln-infinite-loader is used)

maravagyan: Yes, makes sense. I’ll update other places as well.

indigo-frontend/src/app/pages/project/project-list/project-list.component.html
```diff
@@ -26,8 +25,9 @@
         '2xl:col-span-3 md:col-span-2': selectedView == 'grid',
       } | classPicker
     "
+    (enteredViewport)="onInfiniteLoaderEntered()"
```
denis-dzenskevich, line 28: Is it still needed? If yes, shall it be done for other notebooks and experiments too?

denis-dzenskevich (approved):

PR #630, opened 2026-04-15
[ng-620] Fix the experiment item to have icon

smamatov:


indigo-frontend/src/app/pages/experiment/experiment-item/experiment-item.component.html
```diff
@@ -30,6 +30,20 @@
   </div>
 }
 
+        <button
+          class="flex items-center hover:bg-neutral-200 rounded-full cursor-pointer p-1"
+          (click)="toggleMark($event)"
+          matTooltip="Mark experiment as favourite"
```
denis-dzenskevich: We have some mess with how we call it, it's either marked experiments, favorite experiments or starred experiments. But our UI and spec seem to agree on "starred experiments". For the sake of consistency of the UI, please use something like "Add to starred experiments" / "Remove from starred experiments" (in tooltip and aria-label)

indigo-frontend/src/app/pages/experiment/experiment-item/experiment-item.component.ts
```diff
@@ -25,17 +29,51 @@ import { getExperimentStatusBadgeVariant } from '@/core/utils/experiment-status.
...
   @Input() experiment!: ExperimentDetail;
   @Input() variant: 'grid' | 'list' = 'grid';
   @Input() projectId!: string;
   @Input() notebookId!: string;
 
+  isMarked = signal(false);
+
+  constructor() {
+    effect(() => {
+      if (this.experiment) {
```
denis-dzenskevich: I think we never have an experiment card without experiment, so `if` can be removed

indigo-frontend/src/app/pages/experiment/experiment-item/experiment-item.component.ts
```diff
@@ -25,17 +29,51 @@ import { getExperimentStatusBadgeVariant } from '@/core/utils/experiment-status.
...
+        this.isMarked.set(this.experiment.marked ?? false);
+      }
+    });
+  }
+
+  toggleMark(event: Event): void {
```
denis-dzenskevich, line 53: We have "Starred Experiments" section in left pane. Can we reload it when we successfully marked/unmarked the experiment? (by emitting signal or maybe by making ExperimentDetailsService manage the starred experiment list?)

denis-dzenskevich (changes requested):

indigo-frontend/src/app/pages/experiment/experiment-item/experiment-item.component.ts
smamatov: Please check new implementation

indigo-frontend/src/app/pages/experiment/experiment-item/experiment-item.component.ts
smamatov: Resolved

indigo-frontend/src/app/pages/experiment/experiment-item/experiment-item.component.html
smamatov: Resolved and used `Add to starred experiments` as tool tip value

denis-dzenskevich: Can we change tooltip to "Add ..." or "Remove ..." in the same way as aria-label? It would be strange to have "Add ..." tooltip on already added experiment

indigo-frontend/src/app/pages/experiment/experiment-item/experiment-item.component.ts
```diff
@@ -25,17 +29,49 @@ import { getExperimentStatusBadgeVariant } from '@/core/utils/experiment-status.
...
+    this.isMarked.set(nowMarked);
+
+    const request$ = nowMarked
+      ? this.experimentDetailService.mark(this.experiment.id)
+      : this.experimentDetailService.unmark(this.experiment.id);
+
+    request$
+      .pipe(
+        takeUntilDestroyed(this.destroyRef),
+        catchError(() => {
```
denis-dzenskevich: We have ErrorInterceptor that shows notifications in case of backend errors by using the same catchError pipe. It will _probably_ be registered before this one so will be called anyway, but for the sake of clarity, can you please move this code to subscribe? (or maybe rethrow error from catchError)

indigo-frontend/src/core/components/layout/partials/sidebar/starred-experiments/starred-experiments.component.ts
```diff
@@ -33,10 +35,15 @@ export class StarredExperimentsComponent implements OnInit, OnDestroy {
...
   ngOnInit(): void {
     this.fetchMarkedExperiments();
+    this.experimentDetailService.markedChanged$
+      .pipe(takeUntil(this.destroy$))
+      .subscribe(() => this.fetchMarkedExperiments());
   }
 
   private fetchMarkedExperiments(): void {
-    this.loading = true;
+    if (this.experiments.length === 0) {
```
denis-dzenskevich: I guess it won't work as expected when experiments are loaded but empty? Maybe we can have experiments initially null instead of empty list?

denis-dzenskevich (changes requested):

indigo-frontend/src/core/components/layout/partials/sidebar/starred-experiments/starred-experiments.component.ts
smamatov: It is empty (experiments: ExperimentDetail[] = [];)  


https://github.com/epam/Indigo-ELN-v.-2.0/blob/0172055cb1578519aaf14538b56f433d436daa37/indigo-frontend/src/core/components/layout/partials/sidebar/starred-experiments/starred-experiments.component.ts#L29

denis-dzenskevich: Yes, can we set experiments to null there instead of []?

denis-dzenskevich (changes requested):

indigo-frontend/src/app/pages/experiment/experiment-item/experiment-item.component.html
smamatov: Completed

indigo-frontend/src/app/pages/experiment/experiment-item/experiment-item.component.html
```diff
@@ -30,6 +30,20 @@
   </div>
 }
 
+        <button
+          class="flex items-center hover:bg-neutral-200 rounded-full cursor-pointer p-1"
+          (click)="toggleMark($event)"
+          matTooltip="Add to starred experiments"
```
denis-dzenskevich, line 36: Tooltip still not fixed

indigo-frontend/src/app/pages/experiment/experiment-item/experiment-item.component.ts
```diff
@@ -25,17 +28,47 @@ import { getExperimentStatusBadgeVariant } from '@/core/utils/experiment-status.
...
 
   @Input() experiment!: ExperimentDetail;
   @Input() variant: 'grid' | 'list' = 'grid';
   @Input() projectId!: string;
   @Input() notebookId!: string;
 
+  isMarked = signal(false);
+
+  constructor() {
+    effect(() => {
```
denis-dzenskevich, line 48: Why effect is needed? You don't read any signals inside it

indigo-frontend/src/core/components/layout/partials/sidebar/starred-experiments/starred-experiments.component.ts
```diff
@@ -33,15 +35,20 @@ export class StarredExperimentsComponent implements OnInit, OnDestroy {
...
   ngOnInit(): void {
     this.fetchMarkedExperiments();
+    this.experimentDetailService.markedChanged$
+      .pipe(takeUntil(this.destroy$))
+      .subscribe(() => this.fetchMarkedExperiments());
   }
 
   private fetchMarkedExperiments(): void {
-    this.loading = true;
+    if (!this.experiments?.length) {
```
denis-dzenskevich, line 44: If the intent was to only show "loading" the first time, it guess it won't work as expected when there are no experiments? Shall we only check for `this.experiments == null`?

denis-dzenskevich (changes requested):

denis-dzenskevich:
Merged via #686

Issue #553, opened 2026-02-19
[FE] Adjust .prettierrc to allow longer lines


denis-dzenskevich:
Currently prettier splits expressions into multiple lines too aggressively, making expressions harder to read. I suggest to increase allowed line width to 120 or similar

Also, will need to reformat all frontend sources to match the updated style

PR #632, opened 2026-04-18
[FE] Adjust .prettierrc to allow longer lines #553

denis-dzenskevich:
- updated .prettierrc to allow 120-long lines (reformatted files will be in subsequent PR)
- added Lefthook to run prettier on commit
- also added type checking on commit and fixed its remarks
- fixed product-batch-summary-table.component.html after previous merged


.lefthook.yml
```diff
@@ -0,0 +1,11 @@
+pre-commit:
+  jobs:
+    - name: prettier
+      glob: "indigo-frontend/src/**/*.{ts,html,scss,json}"
```
rmonteseba, line 4: I tested this with a real commit on a throwaway branch. Lefthook reported prettier (skip) no files for inspection, which means the Prettier pre-commit job was skipping some staged frontend files instead of formatting them. From what I can tell, this comes from the glob pattern: it matched files in nested folders under src, but not files located directly in src. Is that expected? If so, is there any reason not to also include `indigo-frontend/src/*.{ts,html,scss,json}`?

One additional omission is `src/proxy.conf.js`, which is still excluded because the glob does not include `.js`. Frontend config files outside src also remain excluded by design?.

PR #638, opened 2026-04-22
313 breadcrumbs component

maravagyan:


indigo-frontend/src/core/components/breadcrumbs/breadcrumbs.component.scss
```diff
@@ -0,0 +1,28 @@
+:host {
+  display: block;
+}
+
+.breadcrumbs {
```
denis-dzenskevich: In this project we typically use Tailwind classes instead of CSS. Can you please switch to classes where possible? If some styles is not possible to do or more complex with Tailwind, we can still use CSS

indigo-frontend/src/core/components/breadcrumbs/breadcrumbs.component.ts
```diff
@@ -0,0 +1,283 @@
...
+        this.buildBreadcrumbs();
+      });
+
+    this.buildBreadcrumbs();
+  }
+
+  private buildBreadcrumbs(): void {
+    const requestId = ++this.breadcrumbRequestId;
+
+    if (!this.router.url.startsWith('/projects')) {
```
denis-dzenskevich: Can we investigate a slightly different approach? Our breadcrumbs component is embedded into each relevant page, so we can pass the information from the page to breadcrumbs (e.g. a notebook page can pass {projectID, notebookID} or something similar)

First, it will probably be simpler than parsing the route?

Second, it will probably save us on server calls? Right now, when you open experiment page, you need to wait for 3 additional requests, and because of this breadcrumbs are shows after a delay

Ideally we'd have everything we need without additional requests. E.g. on experiment page we already have experiment name loaded by experiment-service, it will already have experimentName. If needed, I can add also notebookName and projectName (and also projectName to Notebook). It's maybe not very clear as we add data from another entity, but it will be a better UX

indigo-frontend/src/core/components/breadcrumbs/breadcrumbs.component.ts
```diff
@@ -0,0 +1,283 @@
...
+      this.items = [];
+      console.log('[Breadcrumbs] items:', this.items);
```
denis-dzenskevich: Please remove logging in the final version

indigo-frontend/src/app/pages/notebook/notebook-detail/notebook-detail.component.ts
```diff
@@ -7,12 +7,19 @@ import { NotebookService } from '@core/services/notebook/notebook.service';
...
 import { MatDialog } from '@angular/material/dialog';
 import { ProjectTabButtonComponent } from '@pages/project/project-tab-button/project-tab-button.component';
+import { BreadcrumbsComponent } from '@/core/components/breadcrumbs/breadcrumbs.component';
 
 @Component({
   selector: 'eln-notebook-detail',
   templateUrl: './notebook-detail.component.html',
   standalone: true,
-  imports: [RouterOutlet, ProjectTabButtonComponent, ButtonComponent, CardComponent],
+  imports: [
```
denis-dzenskevich, line 18: I've fixed the prettier hooks, please pull changes from 3.0 and run `npm run prettier` (can be done after everything else)

denis-dzenskevich (changes requested):

indigo-frontend/src/app/pages/notebook/notebook-detail/notebook-detail.component.ts
maravagyan: done

denis-dzenskevich (commented):
Can you please update from 3.0 branch, resolve merge conflicts and run `npm run prettier` again?

I like that we have a generic breadcrumbs that can be extended to support other paths (Templates for example). Though it may not be very convenient with the next task in this story: to have dropdowns near Project/Notebook/Experiment parts, so we will probably have to make breadcrumbs component support several *types* of path item, where breadcrumbs component will know how to show dropdown for each of supported item types

Something like
```typescript
export interface BreadcrumbAllProjectsItem {
  type: 'AllProjects';
}

export interface BreadcrumbProjectItem {
  type: 'Project';
  projectId: UUID;
  projectName: string;
}

...

export interface BreadcrumbItem =
  | BreadcrumbAllProjectsItem
  | BreadcrumbProjectItem
  ...;
```

But I think we can leave as is and get back to it when implementing dropdowns

denis-dzenskevich (approved):

PR #640, opened 2026-04-24
#296 backend auth via keycloak

m-markau:
- Enable authentication for backend
- Fix frontend container build
- Migrate from Traefik to Nginx


backend/eln/eln-core/build.gradle.kts
```diff
@@ -7,6 +7,7 @@ plugins {
 dependencies {
     api(project(":eln:eln-api"))
     api(project(":reports:reports-api"))
+    api("io.quarkus:quarkus-oidc")
```
denis-dzenskevich: It's already present in :common:common-service, so should be already available for eln-service

m-markau: This dependency is mandatory. Without it backend services are failing to authenticate. This is because eln-core does not depend on common-service.

 Here's the dependency graph:                                                                                                                                                                                                        
                                                                                                                                                                                                                                      
 common:common  ──► eln:eln-api ──► eln:eln-core   ← what gets packaged & run                                                                                                                                                        
 common:common  ──► reports:reports-api ──┘                                                                                                                                                                                          
 common:common  ──► common:common-service ──► eln:eln-service                                                                                                                                                                        
                                            ──► reports:reports-service                                                                                                                                                               
                                            ──► signature:signature-service

denis-dzenskevich: You don't run eln-core in the container, you run eln-service. eln-core is meant to be a common code between eln-core and eln-service, not a final artifact

As currently written, OIDC is added to lambda, which only needs JWT token without OIDC

m-markau: Switched the build to target eln-service as the runnable Quarkus module. Changes:
                                                                                                                                                                                                                                      
  1. backend/settings.gradle.kts — uncommented include("eln:eln-service") so Gradle picks up the module.                                                                                                                              
  2. backend/eln/eln-core/build.gradle.kts — removed id("io.quarkus") plugin and the QuarkusDev task block. eln-core is now a plain java-library. This also avoids accidentally producing a second runnable Quarkus app and the cross-module confusion that came with two modules both applying the plugin.                                                                                                                                                         
  3. backend/eln/eln-core/Dockerfile — deleted (no longer a deployable artifact).
  4. deployment-compose/docker-compose.yml — eln-core service replaced with eln-service, build context pointed at backend/eln/eln-service, dockerfile set to Dockerfile.jvm. nginx depends_on updated to eln-service.                                                                      
  5. backend/eln/eln-service/src/main/resources/application.properties — flipped quarkus.native.enabled=true → false and removed quarkus.package.jar.enabled=false. Native image builds were taking ~10 min and ~12 GB RAM on every local docker build. JVM fast-jar (default) builds in seconds and is what Dockerfile.jvm consumes (build/quarkus-app/). The native path stays available via eln-lambda for AWS deployments.

denis-dzenskevich (approved):

backend/eln/eln-core/build.gradle.kts
denis-dzenskevich: Merged. Thank you!

Issue #585, opened 2026-03-10
[BE] Validate dictionary item is correct when setting dictionary item


denis-dzenskevich:
We have a single entity/class/table for all dictionaries (as we plan to allow user-defined dictionaries in the future). It means that without extra handling, it's possible to set an item of a wrong dictionary (e.g. it's possible to invoke SetOutputHandlingPrecautions mutation with the value of "Storage Instruction" dictionary instead of "Handling Precautions" dictionary)

- Validate places where DictionaryItemRef or DictionaryItemEntity is set (mostly in experiment model, but also ProjectEntity/NotebookEntity/ExperimentEntity have dictionary fields)
- Validate that item being set is:
    - belongs to a correct dictionary
    - active and not deleted
        - there is an "undo" functionality where model is reverted, and it's possible that undo will bring back a now deleted item, but we'll handle it later
- Maybe add a helper function to validate/set a dictionary item and use it everywhere?
    - but keep in mind that there are DictionaryItemEntity (used in entity classes) and DictionaryItemRef (used in Experiment Model), also we may set a collection of items rather than a single item

See https://kb.epam.com/pages/viewpage.action?spaceKey=EPMLSOPELN&title=Experiment+Data+Model as a reference which field refers to which dictionary

PR #653, opened 2026-04-28
Enforce dictionary item type/active validation and specialize dictionary refs #585

MeriGasparyan:


denis-dzenskevich:
Merged via #657

Issue #374, opened 2025-12-01
[BUG] There is no pop-up window when user is deleted from Project/Notebook level


nadzeyapiatrouskaya-boop:
**Description:**

User removal is currently only possible when selecting the "None" option and then refreshing the page. According to the requirements, the option should be labeled "Remove from Project," and after selecting it, a confirmation pop-up should be displayed.

**Steps to Reproduce:**

1. Open app: https://indigo-eln-dev.test.lifescience.opensource.epam.com/ and login (e.g. as admin)
2. Select any project with added members
3. Navigate to Project Teams section
4. Click on permission option of any user => dropdown is opened
5. Click on "None" option
6. Refresh page and observe the action

**Actual result:** 
User is deleted from Project members

https://github.com/user-attachments/assets/54e34aa3-e37e-47fd-ab11-dcb4608aea68
_Note! Reproduced on Notebook level_

**Expected result:** 
Requirements (AC N 6): https://kb.epam.com/display/EPMLSOPELN/Manage+Projects#ManageProjects-ManageProjectMembers

<img width="1581" height="965" alt="Image" src="https://github.com/user-attachments/assets/5c47cd52-be72-42ec-98a1-204dc5d22b8f" />

Mockups: 
https://www.figma.com/design/RxcXcjpgLOKlORlsWuJ4dR/%F0%9F%8C%9F-Indigo-Designs?node-id=3417-46806&t=zrOpssHA2q0AFKMB-0

PR #685, opened 2026-05-27
There is no pop-up window when user is deleted from Project/Notebook level #374

IamBigBobby:
task: https://github.com/epam/Indigo-ELN-v.-2.0/issues/374


denis-dzenskevich:
Obsolete

PR #702, opened 2026-06-11
fix: migrate material buttons to eln-buttons

aamialkovich:
image
Proposal to update buttons in experiment, so they will look corresponding to mockups.

### DIFF
--- a/indigo-frontend/src/app/pages/experiment/experiment-actions/experiment-actions.component.html
+++ b/indigo-frontend/src/app/pages/experiment/experiment-actions/experiment-actions.component.html
@@ -33,16 +33,14 @@
         !button.inDetailsMenu &&
         (button.allowedStatuses == null || button.allowedStatuses.includes(experiment()?.status))
       ) {
-        <button
-          mat-flat-button
-          [color]="button.color"
-          class="flex items-center gap-1"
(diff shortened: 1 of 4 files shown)


indigo-frontend/src/app/pages/experiment/experiment-actions/experiment-actions.component.html
```diff
@@ -33,16 +33,14 @@
         !button.inDetailsMenu &&
         (button.allowedStatuses == null || button.allowedStatuses.includes(experiment()?.status))
       ) {
-        <button
-          mat-flat-button
-          [color]="button.color"
-          class="flex items-center gap-1"
+        <eln-button
+          [variant]="getButtonVariant(button.color)"
```
denis-dzenskevich: Shall we replace button.color with button.variant directly, and remove the mapping? It's the only place where it's used

denis-dzenskevich (approved):

indigo-frontend/src/app/pages/experiment/experiment-actions/experiment-actions.component.html
aamialkovich: good idea

Issue #697, opened 2026-06-10
[BE] Report Error functionality


denis-dzenskevich:
Add a "report error" method that will save information about incident in a JSON file (directory is configured in application.properties file)
- incident time (current time on a backend)
- user (logged in user)
- text description (passed as a parameter)
- optionally ExperimentSnapshot (experiment passed as optional parameter)
- optionally Mutation (passed as optional parameter)
- optionally attachment (e.g. screenshot)

PR #707, opened 2026-06-15
Add a report API that records current backend state #697

MeriGasparyan:


backend/eln/eln-core/src/main/resources/META-INF/microprofile-config.properties
```diff
@@ -1,6 +1,9 @@
 quarkus.smallrye-openapi.path=/openapi/eln
 quarkus.swagger-ui.path=/swagger/eln
 
+eln.incident.directory=incidents
+%devtest.eln.incident.directory=${user.home}/eln-incidents
```
denis-dzenskevich: Can we use something like "build/incidents" so user home is not cluttered? Also, it should be safer for CI/CD builds (not sure about our AWS CodeBuild, but on Jenkins home directory may not be writeable)

Also, in this way you don't need `clearDirectory` method - it will be cleaned up with gradle clean (or just stay there if you don't use gradle clean, which is also fine)

denis-dzenskevich (approved):

backend/eln/eln-core/src/test/java/com/epam/indigoeln/eln/service/IncidentServiceTest.java
```diff
@@ -0,0 +1,186 @@
...
+                    .sorted()
+                    .toList();
+        }
+    }
+
+    private JsonNode readReport(Path file) throws IOException {
+        return FeignUtil.OBJECT_MAPPER.readTree(file.toFile());
+    }
+
+    private void clearDirectory() throws IOException {
```
denis-dzenskevich: See remark on incident directory

backend/eln/eln-core/src/main/java/com/epam/indigoeln/eln/service/incident/IncidentReportService.java
```diff
@@ -0,0 +1,85 @@
...
+        return objectMapper.readTree(mutationJson);
+    }
+
+    @SneakyThrows
+    @Nullable
+    private String saveAttachment(Path directory, UUID incidentId, @Nullable FileUpload file) {
+        if (file == null) {
+            return null;
+        }
+        String filename = "incident-" + incidentId + "-" + file.fileName();
```
denis-dzenskevich: Claude suggests there is a security issue

file.fileName() is user-controlled. An attacker can submit ../../evil.sh and escape incidentDirectory. Fix: strip to basename only.

```java
  String safeBasename = Path.of(file.fileName()).getFileName().toString();
  String filename = "incident-" + incidentId + "-" + safeBasename;
```

denis-dzenskevich (approved):

PR #712, opened 2026-06-22
[FE] Report Error functionality

rmonteseba:
## Description

- Added a centralized error reporting flow that opens a user-facing dialog when application errors need to be captured.
- Introduced the report dialog UI and supporting textarea field updates to collect clearer issue details and display technical context.
- Wired incident submission through a dedicated frontend service aligned with the new backend endpoint and form-based payload handling.
- Integrated the reporting trigger into global error interception so relevant failures can be surfaced consistently from one place.
- Updated notification behavior and related types so success and error feedback works cleanly with the new reporting flow.

## Demo

https://github.com/user-attachments/assets/253fbb65-ea5c-4cc4-9fbe-7c218a99ab92

### DIFF
--- a/indigo-frontend/src/core/components/common/report-error-dialog/report-error-dialog.component.ts
+++ b/indigo-frontend/src/core/components/common/report-error-dialog/report-error-dialog.component.ts
@@ -9,8 +9,6 @@ import { ReportErrorDialogData, ReportErrorFormValue } from '@core/types/report-
 import { FormlyFieldConfig } from '@ngx-formly/core';
 import { finalize } from 'rxjs';
 
-export type { ReportErrorContext, ReportErrorDialogData, ReportErrorFormValue } from '@core/types/report-error.i';
-
 @Component({
   selector: 'eln-report-error-dialog',
   standalone: true,
--- a/indigo-frontend/src/core/components/common/report-error-dialog/report-error-dialog.component.ts
+++ b/indigo-frontend/src/core/components/common/report-error-dialog/report-error-dialog.component.ts
@@ -1,30 +1,15 @@
 import { CommonModule } from '@angular/common';
-import { Component, inject } from '@angular/core';
-import { FormDialogComponent } from '@core/components/common/form-dialog/form-dialog.component';
+import { Component, inject, signal } from '@angular/core';
 import { MAT_DIALOG_DATA, MatDialogRef } from '@angular/material/dialog';
-import { FormlyFieldConfig } from '@ngx-formly/core';
-import { ReportBugService, ReportErrorTechnicalDetails } from '@core/services/report-bug.service';
+import { FormDialogComponent } from '@core/components/common/form-dialog/form-dialog.component';
 import { NotificationService } from '@core/services/notification/notification.service';
+import { ReportBugService, ReportErrorTechnicalDetails } from '@core/services/report-bug.service';
 import { NotificationType } from '@core/types/notification.i';
+import { ReportErrorDialogData, ReportErrorFormValue } from '@core/types/report-error.i';
+import { FormlyFieldConfig } from '@ngx-formly/core';
 import { finalize } from 'rxjs';
-import { signal } from '@angular/core';
 
-export interface ReportErrorFormValue {
-  title: string;
-  problemDescription: string;
-}
-
-export interface ReportErrorContext {
-  requestURL?: string;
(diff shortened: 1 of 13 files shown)


indigo-frontend/src/core/interceptors/error.interceptor.ts
```diff
@@ -45,3 +63,63 @@ function detectMessage(error: unknown): [string, string] {
   }
   return ['Unknown error', 'Unknown error'];
 }
+
+function buildReportErrorDraft(
+  req: HttpRequest<unknown>,
+  error: HttpErrorResponse,
+  message: string,
+): { initialValue: { title: string; problemDescription: string }; context: ReportErrorContext } {
+  const requestBody = normalizeErrorBody(req.body);
```
denis-dzenskevich, line 69: I don't think we should try too much to have user-friendly representation. Especially for requestBody, it may complicate troubleshooting (e.g. you cannot just paste requestBody to reproduce locally)

I would suggest a minimal normalization (e.g. if string, return as is, otherwise JSON.stringify it)

indigo-frontend/src/core/services/report-error-dialog.service.ts
```diff
@@ -0,0 +1,28 @@
...
+import { MatDialog, MatDialogRef } from '@angular/material/dialog';
+import { ReportErrorDialogComponent } from '@core/components/common/report-error-dialog/report-error-dialog.component';
+import { ReportErrorDialogData, ReportErrorFormValue } from '@core/types/report-error.i';
+
+@Injectable({
+  providedIn: 'root',
+})
+export class ReportErrorDialogService {
+  private dialog = inject(MatDialog);
+  private lastErrorDraft: ReportErrorDialogData | null = null;
```
denis-dzenskevich: Not needed? Since we pass ReportErrorDialogData as action parameter anyway

indigo-frontend/src/core/components/common/report-error-dialog/report-error-dialog.component.ts
```diff
@@ -0,0 +1,127 @@
...
+import { MAT_DIALOG_DATA, MatDialogRef } from '@angular/material/dialog';
+import { FormDialogComponent } from '@core/components/common/form-dialog/form-dialog.component';
+import { NotificationService } from '@core/services/notification/notification.service';
+import { ReportBugService, ReportErrorTechnicalDetails } from '@core/services/report-bug.service';
+import { NotificationType } from '@core/types/notification.i';
+import { ReportErrorDialogData, ReportErrorFormValue } from '@core/types/report-error.i';
+import { FormlyFieldConfig } from '@ngx-formly/core';
+import { finalize } from 'rxjs';
+
+export type { ReportErrorContext, ReportErrorDialogData, ReportErrorFormValue } from '@core/types/report-error.i';
```
denis-dzenskevich: It was probably for the linked HTML file, but it's never used there

denis-dzenskevich (changes requested):

indigo-frontend/src/core/interceptors/error.interceptor.ts
```diff
@@ -45,3 +60,45 @@ function detectMessage(error: unknown): [string, string] {
...
+  }
+
+  if (typeof errorBody === 'string') {
+    return errorBody;
+  }
+
+  try {
+    return JSON.stringify(errorBody);
+  } catch {
+    return 'Unable to serialize error details';
```
denis-dzenskevich, line 99: Please add console log, otherwise in case of error we'd have no clues at all about what has happened

indigo-frontend/src/core/services/report-bug.service.ts
```diff
@@ -0,0 +1,97 @@
...
+    ...(context?.requestBody ? { requestBody: context.requestBody } : {}),
+    ...(context?.responseBody ? { responseBody: context.responseBody } : {}),
+  };
+}
+
+function safePrettyStringify(value: unknown): string {
+  try {
+    return JSON.stringify(value, null, 2);
+  } catch {
+    return 'Unable to serialize value';
```
denis-dzenskevich, line 90: Please add console log, otherwise in case of error we'd have no clues at all about what has happened

denis-dzenskevich (approved):

denis-dzenskevich (approved):

PR #715, opened 2026-06-29
Small frontend fixes

denis-dzenskevich:
- Migrate route parameter extraction from ActivatedRoute to Angular's withComponentInputBinding() + paramsInheritanceStrategy: 'always', so route parameters (including parents) are always available as @Input parameters

- Add ProjectService in the same way as NotebookService; made them singletons to enable reuse data between different components

- In project/notebook details components, trigger load in ProjectService/NotebookService; implement ngOnChanges to catch changes in route parameters

- Fix subscription leak in InfiniteScrollBase.initialize() — store the dataList$ subscription and unsubscribe before re-initializing, resetting list and page state on context switch

- Unified breadcrumbs initialization in project and notebook routes

### DIFF
--- a/indigo-frontend/src/app/pages/project/project-detail/project-detail.component.html
+++ b/indigo-frontend/src/app/pages/project/project-detail/project-detail.component.html
@@ -1,6 +1,6 @@
 <div *projectOverviewWidget="'tab'" class="flex items-center mt-6 overflow-x-auto">
-  <eln-project-tab-button label="Project Info" [routerLink]="projectUrl" />
-  <eln-project-tab-button label="Notebooks" [routerLink]="notebooksUrl" />
+  <eln-project-tab-button label="Project Info" [routerLink]="'/projects/' + this.projectId" />
+  <eln-project-tab-button label="Notebooks" [routerLink]="'/projects/' + this.projectId + '/notebooks'" />
 </div>
 
 <router-outlet />
--- a/indigo-frontend/src/app/pages/project/project-detail/project-detail.component.html
+++ b/indigo-frontend/src/app/pages/project/project-detail/project-detail.component.html
@@ -1,6 +1,6 @@
 <div *projectOverviewWidget="'tab'" class="flex items-center mt-6 overflow-x-auto">
-  <eln-project-tab-button label="Project Info" [routerLink]="'/projects/' + this.projectId" />
-  <eln-project-tab-button label="Notebooks" [routerLink]="'/projects/' + this.projectId + '/notebooks'" />
+  <eln-project-tab-button label="Project Info" [routerLink]="'/projects/' + projectId" />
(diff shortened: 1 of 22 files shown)


indigo-frontend/src/app/pages/project/project-detail/project-detail.component.html
```diff
@@ -1,6 +1,6 @@
 <div *projectOverviewWidget="'tab'" class="flex items-center mt-6 overflow-x-auto">
-  <eln-project-tab-button label="Project Info" [routerLink]="projectUrl" />
-  <eln-project-tab-button label="Notebooks" [routerLink]="notebooksUrl" />
+  <eln-project-tab-button label="Project Info" [routerLink]="'/projects/' + this.projectId" />
```
aamialkovich: Does it work? Seems that we don't need `this` here. Correct me if I'm wrong

indigo-frontend/src/core/components/util/infinite-scroll.base.ts
```diff
@@ -9,13 +9,17 @@ export abstract class InfiniteScrollBase<T> extends PaginatedBase<T> {
   protected override isLoading = false;
 
   private isInfiniteLoaderVisible = false;
+  private dataListSub?: Subscription;
 
   protected override initialize(): void {
+    this.dataListSub?.unsubscribe();
```
aamialkovich, line 18: Maybe we can do unsubscription on component destroy?

aamialkovich (approved):

rmonteseba-epam (changes requested):
Added some comments to consider

indigo-frontend/src/app/pages/project/project-info/project-info.component.html
rmonteseba: Should we use project() instead for keeping consistency?

indigo-frontend/src/core/services/project/project.service.ts
```diff
@@ -0,0 +1,49 @@
...
+      }),
+      finalize(() => {
+        this.isLoading.set(false);
+      }),
+    );
+  }
+
+  refresh() {
+    const id = this.currentId();
+    if (id) this.load(id);
```
rmonteseba: Image

refresh() calls load(id) but never subscribes to it. Since load() returns a cold observable, the request does not run, so the project view can stay stale after closing the edit modal with "refresh". Could we align this with NotebookService and make refresh() actually trigger the reload?

rmonteseba (changes requested):

indigo-frontend/src/core/components/util/infinite-scroll.base.ts
denis-dzenskevich: This is to close subscription to the previous stream when changing filters or sorting. Added takeUntilDestroyed to unsubscribe automatically on destroy, and renamed `initialize` to `reinitialize` to make it clear that it can be called multiple times

aamialkovich (approved):

PR #726, opened 2026-07-08
296 startup issues

m-markau:
Resolve startup issues for docker installation

### DIFF
--- a/backend/database/flyway/src/main/resources/db/data/users.csv
+++ b/backend/database/flyway/src/main/resources/db/data/users.csv
@@ -1,4 +1,2 @@
 id,username,firstName,lastName,displayName,roles
 00000000-0000-0000-0000-000000000001,admin,Administrator,Administrator,Administrator,"Administrators"
-00000000-0000-0000-0000-000000000010,testuser1,Test,User1,Test User1,"Content Editor"
-00000000-0000-0000-0000-000000000011,testuser2,Test,User2,Test User2,"Content Editor"
--- a/backend/database/flyway/src/main/resources/db/data/users.csv
+++ b/backend/database/flyway/src/main/resources/db/data/users.csv
@@ -1,2 +1,4 @@
 id,username,firstName,lastName,displayName,roles
 00000000-0000-0000-0000-000000000001,admin,Administrator,Administrator,Administrator,"Administrators"
+00000000-0000-0000-0000-000000000010,testuser1,Test,User1,Test User1,"Content Editor"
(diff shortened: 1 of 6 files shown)


backend/database/flyway/src/main/resources/db/data/users.csv
```diff
@@ -1,2 +1,4 @@
 id,username,firstName,lastName,displayName,roles
 00000000-0000-0000-0000-000000000001,admin,Administrator,Administrator,Administrator,"Administrators"
+00000000-0000-0000-0000-000000000010,testuser1,Test,User1,Test User1,"Content Editor"
```
denis-dzenskevich: These users will be added to every installation (even production installations). Can we instead create it either manually on test environment, or e.g. in tests (if needed for tests)? For now it looks like they are never user anyway

backend/database/flyway/src/main/resources/db/migration/V1.0.101__signature_service.sql
```diff
@@ -1,7 +1,7 @@
 DROP TABLE Signature_Template_Block;
 DROP TABLE Signature_Template;
 
-ALTER TABLE Experiment DROP COLUMN report_for_signature_id;
+ALTER TABLE Experiment DROP COLUMN report_for_signature_id CASCADE;
```
denis-dzenskevich: On existing database, it will error out. Migration files typically shouldn't be modified, instead, a new migration file should be created

Why do we need cascade here? It seemed to work fine without it on both AWS env and in tests (where database is initialized and all migrations are run)

denis-dzenskevich (changes requested):

backend/database/flyway/src/main/resources/db/migration/V1.0.101__signature_service.sql
m-markau: You're right. I've reverted the CASCADE. Turns out it wasn't actually needed. The failure I hit must have been  from a stale local DB  state.

backend/database/flyway/src/main/resources/db/data/users.csv
m-markau: I added those two test users so I could log into the Frontend as non-admin and exercise different permission sets, if any. I bet no one had been running the docker install yet. That is why it seemed like they have never been used anywere. I've droped them from users.csv . For login to Frontend we will use just admin

denis-dzenskevich (approved):

PR #735, opened 2026-07-15
feat(721): Access Controls for Projects API

rmonteseba:
## PR Summary

- Added `currentPermissions: ApplicationPermission[]` to the `Project` DTO — previously returned by the backend but untyped/unused on the frontend.
- Introduced a reusable permission framework: `PermissionedEntity` interface, `PermissionService.hasPermission(permission, entity?)`, and a `*hasPermission` structural directive (defaults to hidden when data hasn't loaded yet).
- Rule encoded in the framework: creation actions check the user's **global** permissions; actions on an existing entity check that **entity's** `currentPermissions`.
- **Add Project** button (project list) now hidden without `CREATE_PROJECTS`.
- **Edit** button (project info) now hidden without `EDIT_PROJECTS`.
- **Attachments** component: upload control and per-file delete action now hidden without `EDIT_PROJECTS`, via new `requiredPermission`/`entity` inputs (defaults preserve old behavior when unset).
- **Team** component: "add users" control now hidden and ACL level dropdown becomes read-only without `MANAGE_PROJECT_ACCESS`, via the same new inputs.
- Global HTTP error interceptor now shows a friendly message on `403` responses instead of the generic server error text, as a safety net for stale UI/edge cases.
- Framework is entity-agnostic by design — future tickets (e.g. Notebooks, Experiments) only need to add `PermissionedEntity` to their DTO and reuse `PermissionService`/`*hasPermission` without touching `TeamComponent`/`AttachmentsComponent` again.
- No backend changes required, enforcement already exists server-side; this only makes the frontend permission-aware.

image(With permissions view, see non-hidden `add member` and `edit` CTAs)

image (No permissions view, see hidden `add member` and `edit` CTAs)


indigo-frontend/src/core/services/permission/permission.service.ts
```diff
@@ -0,0 +1,36 @@
...
+import { toSignal } from '@angular/core/rxjs-interop';
+import { IdentityService } from '@core/services/identity.service';
+import { PermissionedEntity } from '@core/types/entities/permission.i';
+import { ApplicationPermission } from '@core/types/entities/user.i';
+import { map } from 'rxjs';
+
+/**
+ * Single source of truth for permission checks in the UI.
+ *
+ * Rule of thumb (confirmed across Project and Notebook permission tickets):
```
denis-dzenskevich: Please remove the "confirmed" part, it looks alien in class documentation

indigo-frontend/src/core/directives/has-permission.directive.ts
```diff
@@ -0,0 +1,38 @@
...
+ * - Global permission check (e.g. creating a new entity):
+ *     <button *hasPermission="'CREATE_PROJECTS'">Add Project</button>
+ * - Entity-scoped permission check (uses entity.currentPermissions):
+ *     <button *hasPermission="'EDIT_PROJECTS'; entity: project()">Edit</button>
+ */
+@Directive({
+  selector: '[hasPermission]',
+  standalone: true,
+})
+export class HasPermissionDirective {
```
denis-dzenskevich: I like the idea of a directive, but I have a doubt. If we are asked to disable the control instead of hiding it, do we have a way to do so?

Also, I'm thinking of Experiment screen. In our tables, if user doesn't have EDIT permission or experiment is read-only (when status is other than OPEN/REOPEN) we definitely cannot hide cells and have to disable them

rmonteseba: Moved to disabled approach instead of hidden, lmk your thoughts!

denis-dzenskevich (approved):

PR #747, opened 2026-07-28
#296: add signature container

m-markau:
Add signature container to docker compose deployment


deployment-compose/docker-compose.yml
```diff
@@ -42,6 +43,40 @@ services:
       - "10020:8080"
       - "10021:5005"
 
+  signature-service:
```
denis-dzenskevich, line 47: Please also add "reports" service. See compose-integrationTests-services.yml for details (hopefully we'll soon be able to reuse docker-compose for integration tests). It has no database, only needs to be exposed path "/internalapi/reports" by nginx

denis-dzenskevich (changes requested):

deployment-compose/docker-compose.yml
m-markau: Done

denis-dzenskevich (approved):

Issue #644, opened 2026-04-26
[BUG] There is no status filter (Experiment dashboard)


nadzeyapiatrouskaya-boop:
**Steps to Reproduce:**

1. Open app: https://indigo-eln-dev.test.lifescience.opensource.epam.com/ and login (e.g. as admin)
2. Open Experiment Dashboard
3. Take a look at the filter options

**Actual result:** 
There is no status filter (Experiment dashboard)

<img width="1914" height="1008" alt="Image" src="https://github.com/user-attachments/assets/463b854c-1421-45af-b74b-05d21b6920c5" />

**Expected result:** 
Status filter should be near "My Entities" toggle button

<img width="1585" height="904" alt="Image" src="https://github.com/user-attachments/assets/0c5b6f1a-1696-4e26-8096-b9e40ce2d5d4" />

Mockups: https://www.figma.com/proto/RxcXcjpgLOKlORlsWuJ4dR/%F0%9F%8C%9F-Indigo-Designs?node-id=2126-7760&t=GCXde83TUNz4nV4w-0&scaling=scale-down&content-scaling=fixed&page-id=3193%3A35034&starting-point-node-id=2080%3A6664&show-proto-sidebar=1

PR #751, opened 2026-07-29
Fix status filter #644

alexfil3:


indigo-frontend/src/app/pages/notebook/notebook-experiments-tab/notebook-experiments-tab.component.ts
```diff
@@ -40,19 +43,85 @@ import { ProjectOverviewWidgetDirective } from '@pages/project/projects-overview
...
   selectedView: 'grid' | 'list' = 'grid';
+  headerSortOptions: DropdownMenuItem[] = [];
   @Input() notebookId!: string;
+  headerFilterOptions: CheckboxDropdownItem[] = [];
 
-  ngOnChanges(changes: SimpleChanges): void {
+  ngOnInit() {
+    this.headerSortOptions = this.getSortOptions().map((option) => ({
+      label: `${option.label}`,
+      value: `${option.value}:${option.defaultOrder}`,
```
denis-dzenskevich, line 53: Looks like our frontend codebase is slightly off. On backend, we only have sort options EARLIEST and LATEST, without specifying which field to sort (modeled in this way because mockups have only flat list of sort options, currently only these two)

I will create a separate ticket to update frontend

indigo-frontend/src/core/components/common/checkbox-dropdown/checkbox-dropdown.component.ts
```diff
@@ -0,0 +1,81 @@
...
+  animations: [dropdownAnimation],
+  providers: [
+    {
+      provide: NG_VALUE_ACCESSOR,
+      useExisting: forwardRef(() => CheckboxDropdownComponent),
+      multi: true,
+    },
+  ],
+})
+export class CheckboxDropdownComponent extends DropdownBaseComponent implements AfterViewInit, ControlValueAccessor {
```
denis-dzenskevich, line 21: When collapsed, it doesn't show what is selected. I didn't find the mockups with how expanded filter should look like, maybe we'll reuse existing multiselect component? E.g. EnumSelectComponent, it will show selected items as chips inside of a text box

(hopefully we'll have Storybook at some point and we'll be able to see what all components we have to get rid of possible duplication)

indigo-frontend/src/app/pages/notebook/notebook-experiments-tab/notebook-experiments-tab.component.ts
```diff
@@ -40,19 +43,85 @@ import { ProjectOverviewWidgetDirective } from '@pages/project/projects-overview
...
+  onViewChange(view: string) {
+    this.selectedView = view as 'grid' | 'list';
+  }
+
+  onMyEntitiesOnlyChange(value: boolean) {
+    this.filters['createdByMe'] = value;
+    this.reload();
+  }
+
+  onFilterChange(items: CheckboxDropdownItem[]) {
```
denis-dzenskevich, line 102: It works differently from existing logic used by createdByMe filter and sort. They update query params in page URL and reload pageable with new params

indigo-frontend/src/app/pages/notebook/notebook-experiments-tab/notebook-experiments-tab.component.ts
```diff
@@ -40,19 +43,85 @@ import { ProjectOverviewWidgetDirective } from '@pages/project/projects-overview
...
 
-  ngOnChanges(changes: SimpleChanges): void {
+  ngOnInit() {
+    this.headerSortOptions = this.getSortOptions().map((option) => ({
+      label: `${option.label}`,
+      value: `${option.value}:${option.defaultOrder}`,
+      icon: 'indicon-sort',
+    }));
+
+    this.headerFilterOptions = this.getFilterOptions().map((option) => ({
```
denis-dzenskevich: So it equals to this.getFilterOptions() without change?

denis-dzenskevich (changes requested):

indigo-frontend/src/core/components/common/checkbox-dropdown/checkbox-dropdown.component.ts
alexfil3: I would prefer to go with a new component as I did, because it's probably harder to reuse existing dropdown-menu. And the purpose of dropdown-menu is to select one of many and goal of checkbox-dropdown is to select any of many. So here I will do what you decide, but before taking a decision please take a look at the figma design for this component, here is the [link](https://www.figma.com/design/RxcXcjpgLOKlORlsWuJ4dR/%F0%9F%8C%9F-Indigo-Designs?node-id=11849-6199&p=f&t=r81WMdp43X3SLHW3-0). Preview 
Screenshot 2026-07-29 175629

denis-dzenskevich: The expanded session looks better, but when collapsed, user has no idea what was selected. Looks very user-unfriendly to me

> And the purpose of dropdown-menu is to select one of many and goal of checkbox-dropdown is to select any of many

Yes we have multiselect too

resim

indigo-frontend/src/app/pages/notebook/notebook-experiments-tab/notebook-experiments-tab.component.ts
alexfil3: Updated. reload() now updates URL query params via fetchDataAndUpdateQueryParams(), so filter behavior matches createdByMe and sort

indigo-frontend/src/app/pages/notebook/notebook-experiments-tab/notebook-experiments-tab.component.ts
alexfil3: Yeah, it was a no-op map. I replaced it with this.getFilterOptions() directly

indigo-frontend/src/core/components/common/checkbox-dropdown/checkbox-dropdown.component.ts
alexfil3: In this case should we ask Daria Zhilina to update the design, so we can see how the user will see chosen statuses?

indigo-frontend/src/core/utils/http.util.ts
```diff
@@ -6,9 +6,26 @@ export function getRequestParams(filter: Record<string, unknown>, paging?: Paged
 
   const filters = { ...filter, ...paging };
 
-  Object.keys(filters).forEach(
-    (key) => (typeof filters[key] == 'number' || !!filters[key]) && (params = params.append(key, filters[key])),
-  );
+  Object.keys(filters).forEach((key) => {
+    const value = filters[key];
+
+    if (value === null || value === undefined || value === '') {
```
denis-dzenskevich, line 12: Can we add "false" too? createdByMe filter is false by default, so it will be sent as true or false on each request. Contrary to expected behavior when only enabled filters will affect query parameters. In our case, it's no effect, as createdByMe on backend is effective only when "true"

Alternatively, let createdByMe filter use true/null instead of true/false in filters map, would be more correct

denis-dzenskevich (changes requested):

indigo-frontend/src/core/utils/http.util.ts
alexfil3: Updated with true/null option, added method to paginated.base and reused it in code where we set boolean filter for createdByMe

denis-dzenskevich (approved):

PR #752, opened 2026-08-03
feat(750): ux loader improvement

### LINKED ISSUE #750: [FE] [UX] Provide visual feedback for in-flight operations
When API call is in progress, make UI reflect it (show spinner over button or other relevant components)

- identify where API calls are made
- decide how they should look (e.g. button will be disabled, it's text will transform into spinner or table becomes disabled and shows spinner centered over table viewport)
- create/updated relevant components (based on previous step)
- update API calls to show visual feedback

leydytatgue:
Resolves #750 [FE] [UX] Provide visual feedback for in-flight operations

# Summary
This PR improves the user experience by adding visible loading feedback during in-flight operations in the frontend.

## What changed
* Added loading states to primary actions so users can clearly see when an operation is in progress.
* Updated the shared button component to support spinner-based loading feedback.
* Extended the form dialog component to support loading state on submit actions.
* Added visual loading overlays for search result areas while requests are running.
* Disabled duplicate submission during pending API calls to prevent repeated actions.

### Shared UI components
* `button.component.ts`
* `button.component.html`
* `form-dialog.component.ts`
* `form-dialog.component.html`

### Search flows
* `global-search.component.ts`
* `global-search.component.html`
* `sample-search.component.ts`
* `sample-search.component.html`
* `sample-search-results.component.ts`
* `sample-search-results.component.html`

# Evidence

https://github.com/user-attachments/assets/c543d96f-b71f-40ee-ad9a-a276b45b5b8e

### DIFF
--- a/indigo-frontend/src/app/pages/experiment/sample-search-results/sample-search-results.component.html
+++ b/indigo-frontend/src/app/pages/experiment/sample-search-results/sample-search-results.component.html
@@ -42,7 +42,7 @@
         <eln-button
           variant="grey"
           classList="ml-2 min-w-[42px]"
-          [disabled]="isAddingToExperiment || (loadingSampleKey != null && !isRowLoading(element))"
+          [disabled]="isAddingToExperiment"
           [loading]="isRowLoading(element)"
           (click)="addToExperiment.emit(element)"
         >
--- a/indigo-frontend/src/app/pages/experiment/sample-search-results/sample-search-results.component.html
+++ b/indigo-frontend/src/app/pages/experiment/sample-search-results/sample-search-results.component.html
@@ -40,10 +40,17 @@
       </ng-template>
       <ng-template elnColumnDef="add" displayName="" let-element>
         <button
-          class="ml-2 px-4 py-2 bg-[#f3f3f3] text-black rounded-[5px] cursor-pointer"
+          class="ml-2 px-4 py-2 bg-[#f3f3f3] text-black rounded-[5px] cursor-pointer disabled:opacity-70 disabled:cursor-not-allowed"
+          [disabled]="isAddingToExperiment"
           (click)="addToExperiment.emit(element)"
         >
-          <mat-icon matTooltip="Add to current experiment" class="text-[#3f50b5]">add </mat-icon>
+          @if (isAddingToExperiment) {
+            <span class="inline-flex items-center gap-2">
+              <span class="h-4 w-4 animate-spin rounded-full border-2 border-[#3f50b5] border-t-transparent"></span>
+            </span>
+          } @else {
+            <mat-icon matTooltip="Add to current experiment" class="text-[#3f50b5]">add </mat-icon>
+          }
         </button>
       </ng-template>
 
--- a/indigo-frontend/src/app/pages/experiment/sample-search-results/sample-search-results.component.html
+++ b/indigo-frontend/src/app/pages/experiment/sample-search-results/sample-search-results.component.html
@@ -39,19 +39,15 @@
         }
       </ng-template>
       <ng-template elnColumnDef="add" displayName="" let-element>
-        <button
-          class="ml-2 px-4 py-2 bg-[#f3f3f3] text-black rounded-[5px] cursor-pointer disabled:opacity-70 disabled:cursor-not-allowed"
-          [disabled]="isAddingToExperiment"
+        <eln-button
+          variant="grey"
+          classList="ml-2 min-w-[42px]"
+          [disabled]="isAddingToExperiment || (loadingSampleKey != null && !isRowLoading(element))"
+          [loading]="isRowLoading(element)"
           (click)="addToExperiment.emit(element)"
         >
-          @if (isAddingToExperiment) {
-            <span class="inline-flex items-center gap-2">
-              <span class="h-4 w-4 animate-spin rounded-full border-2 border-[#3f50b5] border-t-transparent"></span>
-            </span>
-          } @else {
-            <mat-icon matTooltip="Add to current experiment" class="text-[#3f50b5]">add </mat-icon>
-          }
-        </button>
+          <mat-icon matTooltip="Add to current experiment" class="text-[#3f50b5]">add </mat-icon>
+        </eln-button>
       </ng-template>
 
       <ng-template #details let-element>
--- a/indigo-frontend/src/app/pages/experiment/sample-search-results/sample-search-results.component.ts
+++ b/indigo-frontend/src/app/pages/experiment/sample-search-results/sample-search-results.component.ts
@@ -16,6 +16,7 @@ import { MatIcon, MatIconModule } from '@angular/material/icon';
 import { MatFormFieldModule } from '@angular/material/form-field';
 import { MatButtonModule } from '@angular/material/button';
 import { ApiService } from '@core/services/api.service';
+import { ButtonComponent } from '@core/components/common/button/button.component';
 
 @Component({
   standalone: true,
@@ -34,12 +35,14 @@ import { ApiService } from '@core/services/api.service';
     MatFormFieldModule,
     MatButtonModule,
     MatIconModule,
+    ButtonComponent,
   ],
   templateUrl: './sample-search-results.component.html',
 })
 export class SampleSearchResultsComponent {
   @Input({ required: true }) loader: SamplesSearchLoader;
   @Input() isAddingToExperiment = false;
+  @Input() loadingSampleKey: string | null = null;
   @Output() addToExperiment = new EventEmitter<Sample>();
 
   destroyRef = inject(DestroyRef);
@@ -52,6 +55,14 @@ export class SampleSearchResultsComponent {
       .subscribe((response) => {
         this.loader.replace((s) => s.id === sample.id, response);
       });
+  }
+
+  getRowKey(sample: Sample): string {
+    return sample.id ?? sample.compoundKey ?? sample.name ?? 'new-sample';
+  }
+
+  isRowLoading(sample: Sample): boolean {
+    return this.loadingSampleKey != null && this.loadingSampleKey === this.getRowKey(sample);
   }
 
   protected readonly encodeURIComponent = encodeURIComponent;
--- a/indigo-frontend/src/app/pages/experiment/sample-search-results/sample-search-results.component.ts
+++ b/indigo-frontend/src/app/pages/experiment/sample-search-results/sample-search-results.component.ts
@@ -39,6 +39,7 @@ import { ApiService } from '@core/services/api.service';
 })
 export class SampleSearchResultsComponent {
   @Input({ required: true }) loader: SamplesSearchLoader;
+  @Input() isAddingToExperiment = false;
   @Output() addToExperiment = new EventEmitter<Sample>();
 
   destroyRef = inject(DestroyRef);
--- a/indigo-frontend/src/app/pages/experiment/sample-search/sample-search.component.html
+++ b/indigo-frontend/src/app/pages/experiment/sample-search/sample-search.component.html
@@ -315,21 +315,17 @@
           </div>
         </section>
 
-        <button
+        <eln-button
           type="button"
-          class="ml-2 px-4 py-2 bg-[#3F50B5] text-white rounded-[5px] cursor-pointer disabled:opacity-70 disabled:cursor-not-allowed"
-          [disabled]="isSearching"
+          variant="blue"
+          classList="ml-2"
+          [disabled]="loader.loading"
+          [loading]="loader.loading"
+          [loadingText]="'Searching...'"
           (click)="performSearch()"
         >
-          @if (isSearching) {
-            <span class="inline-flex items-center gap-2">
-              <span class="h-4 w-4 animate-spin rounded-full border-2 border-white border-t-transparent"></span>
-              Searching...
-            </span>
-          } @else {
-            Search
-          }
-        </button>
+          Search
+        </eln-button>
       </div>
     </div>
   </form>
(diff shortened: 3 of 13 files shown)


indigo-frontend/src/app/pages/experiment/sample-search-results/sample-search-results.component.html
```diff
@@ -40,10 +40,17 @@
       </ng-template>
       <ng-template elnColumnDef="add" displayName="" let-element>
         <button
-          class="ml-2 px-4 py-2 bg-[#f3f3f3] text-black rounded-[5px] cursor-pointer"
+          class="ml-2 px-4 py-2 bg-[#f3f3f3] text-black rounded-[5px] cursor-pointer disabled:opacity-70 disabled:cursor-not-allowed"
+          [disabled]="isAddingToExperiment"
```
denis-dzenskevich: Can we reuse eln-button with loading so we don't need to write conditionals afterwards?

indigo-frontend/src/app/pages/experiment/sample-search/sample-search.component.ts
```diff
@@ -187,6 +189,10 @@ export class SampleSearchComponent implements OnInit {
   }
 
   performSearch() {
+    if (this.isSearching) {
```
denis-dzenskevich: Please reuse eln-button if possible and in eln-button, make sure disabled or loading state prevents clicks (via both mouse and keyboard)

indigo-frontend/src/app/pages/experiment/sample-search/sample-search.component.html
```diff
@@ -332,5 +340,9 @@ <h2 class="font-semibold text-lg mb-4">{{ title }}</h2>
     }
   </div>
 
-  <eln-sample-search-results [loader]="loader" (addToExperiment)="addToExperiment($event)" />
+  <eln-sample-search-results
```
denis-dzenskevich, line 338: Here it looks a little odd because all rows show spinner at the same time. Is it possible to show spinner only for clicked row, and disable all others?

indigo-frontend/src/core/components/common/button/button.component.html
```diff
@@ -1,3 +1,18 @@
...
+  [type]="type"
+  [disabled]="disabled || loading"
+  [class]="variant | buttonVariant: classList"
+  [attr.aria-label]="ariaLabel"
+  [attr.aria-busy]="loading"
+>
+  @if (loading) {
+    <span class="inline-flex items-center gap-2">
+      <span class="h-4 w-4 animate-spin rounded-full border-2 border-current border-t-transparent"></span>
+      @if (loadingText) {
```
denis-dzenskevich: A side effect of having alternative text is button resizing when switching from normal/loading phase. Maybe we can still show the old text (to preserve old size), but also put centered spinner that obscures the text (maybe make text transparent)?

indigo-frontend/src/app/pages/experiment/sample-search/sample-search.component.ts
```diff
@@ -115,6 +115,8 @@ export class SampleSearchComponent implements OnInit {
   notificationService = inject(NotificationService);
 
   title = 'Add Material';
+  isSearching = false;
```
denis-dzenskevich: Seems never resets to false?

indigo-frontend/src/app/pages/experiment/sample-search/sample-search.component.ts
```diff
@@ -271,12 +288,18 @@ export class SampleSearchComponent implements OnInit {
...
+      next: () => {
+        this.isAddingToExperiment = false;
+        this.notificationService.notify({
+          type: NotificationType.Info,
+          message: 'Model updated with new sample',
+          isInline: false,
+        });
+      },
+      error: () => {
+        this.isAddingToExperiment = false;
```
denis-dzenskevich: I know it's how the code is written, but it starts getting harder to track. E.g. we have to remember to reset isAddingToExperiment in three places. Can you please rewrite to async pipeline?

Something like this or similar

```typescript
this.apiService.request(...)
  .pipe(
    flatMap(this.doAddToExperiment),
    tap(() => /* show notification */),
    finalize(() => this.isAddingToExperiment = false),
  )
```

indigo-frontend/src/app/pages/experiment/sample-search-results/sample-search-results.component.ts
```diff
@@ -39,6 +39,7 @@ import { ApiService } from '@core/services/api.service';
 })
 export class SampleSearchResultsComponent {
   @Input({ required: true }) loader: SamplesSearchLoader;
+  @Input() isAddingToExperiment = false;
```
denis-dzenskevich, line 44: Never resets?

indigo-frontend/src/app/pages/search/global-search/global-search.component.ts
```diff
@@ -225,6 +231,10 @@ export class GlobalSearchComponent implements OnInit, AfterViewInit {
     this.dialogRef.close();
   }
 
+  onSearchCompleted(): void {
```
denis-dzenskevich: Never called?

denis-dzenskevich (changes requested):

indigo-frontend/src/app/pages/experiment/sample-search/sample-search.component.ts
denis-dzenskevich: Also, maybe we can use a custom pipe operation that will handle the flag for us?

```typescript
export function Loading(signal: WritableSignal<boolean>) {
  return <T>(source: Observable<T>): Observable<T> => {
    signal.set(true);
    return source.pipe(finalize(() => signal.set(false)));
  };
}

// simple case

isAddingToExperiment = signal(false);

this.service.request(...)
  .pipe(withLoading(this.isAddingToExperiment))

// our case

this.apiService.request(...)
  .pipe(
    flatMap(this.doAddToExperiment),
    tap(() => /* show notification */),
    withLoading(this.isAddingToExperiment),
  )
```

Not sure about naming, maybe trackLoading or indicateLoading

indigo-frontend/src/app/pages/experiment/sample-search-results/sample-search-results.component.html
leydytatgue: Loading is in isRowLoading (element) and a shared disabled condition, so the per-row spinner is handled by the reusable button component and we avoid extra manual conditionals.

indigo-frontend/src/app/pages/experiment/sample-search/sample-search.component.ts
leydytatgue: I switched the row action to `eln-button` and kept the loading/disabled state on the button itself. Since `eln-button` uses the native `disabled` attribute when `loading` or `disabled` is true, mouse and keyboard interactions are blocked during load.

indigo-frontend/src/app/pages/experiment/sample-search/sample-search.component.html
leydytatgue: I updated the row action so only the clicked row shows `loading`, and all other row add buttons are disabled while that request is in flight.

indigo-frontend/src/core/components/common/button/button.component.html
leydytatgue: I updated `eln-button` so the original label stays rendered and is only hidden with opacity when `loading` is active, while a centered spinner is overlaid. This preserves the button size during the loading phase and avoids the text resizing jump.

indigo-frontend/src/app/pages/experiment/sample-search/sample-search.component.ts
leydytatgue: That `isSearching `line was from an earlier draft and is no longer used in the current implementation. Now the search guard relies on `[this.loader?.loading]`, so loading resets correctly when the loader completes.

indigo-frontend/src/app/pages/experiment/sample-search/sample-search.component.ts
leydytatgue: I rewrote `addToExperiment()` to use the `withLoading(...)` RxJS pipe helper, so the loading flag is set and cleared inside the observable pipeline instead of manually in multiple places. This keeps the async flow centralized and avoids separate reset logic in the subscription.

indigo-frontend/src/app/pages/experiment/sample-search-results/sample-search-results.component.ts
leydytatgue: `isAddingToExperiment` is only an input on `sample-search-results`; the actual flag is managed in the parent sample-search.component.ts. In `addToExperiment()` the `withLoading(...)` helper clears `isAddingToExperiment` on finalize, so the parent resets it when the request completes.

indigo-frontend/src/app/pages/search/global-search/global-search.component.ts
leydytatgue: That method was dead code and not referenced anywhere, so it has been removed. The current search flow now uses `loader.loading` directly instead.

indigo-frontend/src/app/pages/experiment/sample-search-results/sample-search-results.component.html
```diff
@@ -39,12 +39,15 @@
         }
       </ng-template>
       <ng-template elnColumnDef="add" displayName="" let-element>
-        <button
-          class="ml-2 px-4 py-2 bg-[#f3f3f3] text-black rounded-[5px] cursor-pointer"
+        <eln-button
+          variant="grey"
+          classList="ml-2 min-w-[42px]"
+          [disabled]="isAddingToExperiment || (loadingSampleKey != null && !isRowLoading(element))"
```
denis-dzenskevich: Maybe we can simplify it to just "isAddingToExperiment"? Button will be both disabled=true and loading=true, which is effectively the same as loading=true, right?

indigo-frontend/src/app/pages/experiment/sample-search/sample-search.component.ts
```diff
@@ -187,6 +193,10 @@ export class SampleSearchComponent implements OnInit {
   }
 
   performSearch() {
+    if (this.loader?.loading) {
```
denis-dzenskevich: Probably not needed as we already disabled the button? Actually, it doesn't hurt anything, if you think it's more clear, let it stay

indigo-frontend/src/app/pages/experiment/sample-search/sample-search.component.ts
```diff
@@ -256,28 +266,52 @@ export class SampleSearchComponent implements OnInit {
   }
 
   addToExperiment(sample: Sample) {
-    if (!sample.id) {
-      this.apiService.request<Sample>('post', '/samples/importFromSearch', sample).subscribe((response) => {
-        this.doAddToExperiment(response.id);
-      });
-    } else {
-      this.doAddToExperiment(sample.id);
+    if (this.isAddingToExperiment) {
```
denis-dzenskevich: Same as above, can be removed as we already disabled the button

indigo-frontend/src/app/pages/search/global-search/global-search.component.ts
```diff
@@ -169,6 +169,10 @@ export class GlobalSearchComponent implements OnInit, AfterViewInit {
   }
 
   performSearch() {
+    if (this.loader?.loading) {
```
denis-dzenskevich: Same as above, can be removed as we already disabled the button

indigo-frontend/src/core/components/common/button/button.component.ts
```diff
@@ -12,6 +12,8 @@ import { ButtonVariants } from './button.variant';
 export class ButtonComponent {
   @Input() variant: ButtonVariants['variant'] = 'green';
   @Input() disabled = false;
+  @Input() loading = false;
+  @Input() loadingText = 'Loading...';
```
denis-dzenskevich: loadingText can be removed now?

denis-dzenskevich (changes requested):

indigo-frontend/src/core/components/common/button/button.component.ts
leydytatgue: Removed loadingText. eln-button no longer exposes that input and I also removed it from eln-file-upload to reduce the API and maintain consistent behavior.

indigo-frontend/src/app/pages/search/global-search/global-search.component.ts
leydytatgue: "I removed the redundant check. The button is already disabled while loading, so the early check wasn't necessary. Row loading still works via loadingSampleKey / isRowLoading()

indigo-frontend/src/app/pages/experiment/sample-search/sample-search.component.ts
leydytatgue: Removed the redundant guard. The add/search buttons are disabled while loading and loading is managed centrally via withLoading(...), so the early if check is unnecessary. Per-row spinner still works via loadingSampleKey/isRowLoading().

indigo-frontend/src/app/pages/experiment/sample-search/sample-search.component.ts
leydytatgue: I removed the redundant loader.loading guard. The button is already disabled while loading

indigo-frontend/src/app/pages/experiment/sample-search-results/sample-search-results.component.html
leydytatgue: simplified to use `isAddingToExperiment` only. The per-row spinner still comes from isRowLoading(element), so the clicked row shows loading while `isAddingToExperiment `disables the others. I also removed the early guard in `addToExperiment() `and centralized loading with [withLoading(...)(url).

denis-dzenskevich (approved):

PR #754, opened 2026-08-04
#450 Sonar Deployment

denis-dzenskevich:


backend/eln/eln-api/src/main/java/com/epam/indigoeln/reaction/model/CompoundRef.java
```diff
@@ -49,7 +50,7 @@ public sealed interface CompoundRef permits CompoundRef.StoredOrVirtual, Compoun
 
     @Nullable
     @SuppressWarnings("unused") // used on frontend
-    EnteredValue<NoUnit> getExactMass();
+    EnteredValue<@NonNull NoUnit> getExactMass();
```
denis-dzenskevich: I rolled back this change, shouldn't be needed, as we use @NullsMarked on package level, so everything is non-null unless annotated @Nullable. If it triggers Sonar warning, need to take a closer look

PR #756, opened 2026-08-10
feat(722): [FE] Access Control for Notebooks API

rmonteseba:
- Added explicit handling for global and entity-scoped permissions.
- Applied notebook creation permissions to all notebook creation entry points.
- Restricted notebook editing actions based on entity-level permissions.
- Added permission checks for notebook attachment upload, download, and deletion.
- Restricted notebook team management actions based on access permissions.
- Kept unauthorized actions visible but disabled, with defensive checks to prevent unauthorized requests.

image

### DIFF
--- a/indigo-frontend/src/app/pages/project/project-info/project-info.component.ts
+++ b/indigo-frontend/src/app/pages/project/project-info/project-info.component.ts
@@ -53,9 +53,6 @@ export class ProjectInfoComponent {
     const project = this.project();
     return this.permissionService.hasEntityPermission(ApplicationPermission.EDIT_PROJECTS, project);
   });
-  canCreateNotebook = computed(() =>
-    this.permissionService.hasGlobalPermission(ApplicationPermission.CREATE_NOTEBOOKS),
-  );
 
   projectTeamConfig: TeamComponentConfig = {
     buildAccessEndpoint: (id: string) => `projects/${id}/access`,
@@ -81,7 +78,7 @@ export class ProjectInfoComponent {
     }
 
     if (mode === projectInfoModalEnum.NOTEBOOK) {
-      if (!this.canCreateNotebook()) return;
+      if (!this.projectService.canCreateNotebook()) return;
 
       ref = this.dialog.open(NotebookAddComponent);
       (ref.componentInstance as NotebookAddComponent).projectId = this.projectId;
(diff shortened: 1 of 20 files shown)


indigo-frontend/src/app/pages/project/project-info/project-info.component.ts
```diff
@@ -51,8 +51,11 @@ export class ProjectInfoComponent {
   hasError = this.projectService.hasError;
   canEditProject = computed(() => {
     const project = this.project();
-    return project ? this.permissionService.hasPermission(ApplicationPermission.EDIT_PROJECTS, project) : false;
+    return this.permissionService.hasEntityPermission(ApplicationPermission.EDIT_PROJECTS, project);
   });
+  canCreateNotebook = computed(() =>
```
denis-dzenskevich, line 56: CREATE_NOTEBOOKS is entity-level permission on Project rather than a global permission, see https://kb.epam.com/spaces/EPMLSOPELN/pages/2861004595/Action+Permissions

Plus, method duplicated 3 times. Maybe move it to e.g. ProjectService?

indigo-frontend/src/core/components/common/attachment/attachment.component.html
```diff
@@ -18,7 +18,7 @@
         <mat-icon>more_horiz</mat-icon>
       </button>
       <mat-menu #menu="matMenu">
-        <button mat-menu-item (click)="downloadAttachment()">
+        <button mat-menu-item [disabled]="!canDownload()" (click)="canDownload() && downloadAttachment()">
```
denis-dzenskevich, line 21: Double-checking of canDownload() in click seems excessive?

indigo-frontend/src/core/components/common/attachment/attachment.component.ts
```diff
@@ -45,12 +46,16 @@ export class AttachmentComponent {
...
   downloadAttachment() {
+    if (!this.canDownload()) return;
+
     this.downloadService
       .download('get', `${this.baseURL()}/attachments/${this.attachment().id}`, 'attachment')
       .subscribe();
   }
 
   deleteAttachment() {
+    if (!this.canDelete()) return;
```
denis-dzenskevich, line 57: Debatable, but I would remove these checks to simplify code slightly. These operations should already be disabled via controls

If this is to protect from malicious users who can flip the disabled property via Dev Tools, they can send the request directly as well. Backend will verify permissions again anyway

(if you think it's more clear this way, let it stay)

denis-dzenskevich (changes requested):

denis-dzenskevich (approved):

